Issue #105, opened 2023-01-31
Support Smithy 2.0 features


alex-chew:
[Smithy 2.0](https://smithy.io/2.0/) is now available. Polymorph is built on Smithy 1.0 and should be updated.

There is a migration guide for upgrading from version 1.0 of the Smithy IDL to version 2.0: https://smithy.io/2.0/guides/migrating-idl-1-to-2.html

Most of the changes seem to only affect the IDL, and not the semantic model, but there are (at least) these notable exceptions:

* [The `@box` trait is removed](https://smithy.io/2.0/guides/migrating-idl-1-to-2.html#replace-the-box-trait), and [primitive root-level shapes have no implicit default](https://smithy.io/2.0/guides/migrating-idl-1-to-2.html#converting-primitive-root-level-shapes-from-1-0)
* [The `set` shape is deprecated](https://smithy.io/2.0/guides/migrating-idl-1-to-2.html#convert-set-shapes-to-list-shapes) in favor of a `list` shape with the `@uniqueItems` trait

PR #202, opened 2023-03-24
feat: support Smithy 2.0

alex-chew:
*Issue #, if available:* #105

*Description of changes:* Implements Smithy 2.0 support. Concretely, this updates Smithy dependencies to the latest version (1.28.1 as of this writing) which can read Smithy IDL 2.0 models.

Most of smithy-dafny's code generation logic is unchanged because the underlying Smithy semantic model has not undergone any breaking changes. The exception is that supporting Smithy 2.0's new `enum` shape (as opposed to `@enum`-attached `string` shapes) requires changes to shape discovery and generated symbol names. Thankfully the Smithy tooling attaches `SyntheticEnumTrait (extends EnumTrait)` to each Smithy 2.0 `enum` shape, so that the codegen logic can inspect `enum` shapes just like the old `@enum string` shapes.

Notes for reviewers:

* The `SimpleEnumV2` test model is mostly copied from the `SimpleEnum` test model, with names changed to include "V2". I'd recommend only skimming over these files.
* The most important changes are those made in the codegen logic. In particular, I'm less familiar with the Java codegen structure, so I'd appreciate extra attention to it (cc @texastony). 

### DIFF
--- a/.github/workflows/test_models_dafny_verification.yml
+++ b/.github/workflows/test_models_dafny_verification.yml
@@ -49,7 +49,7 @@ jobs:
           # TestModels/SimpleTypes/SimpleByte,
           TestModels/SimpleTypes/SimpleDouble,          
           TestModels/SimpleTypes/SimpleEnum,
-          # TestModels/SimpleTypes/SimpleEnumV2,
+          TestModels/SimpleTypes/SimpleEnumV2,
           # TestModels/SimpleTypes/SimpleFloat,
           TestModels/SimpleTypes/SimpleInteger,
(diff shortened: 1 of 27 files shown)


robin-aws (changes requested):
Looks great! I always love flipping "will work later" to "does work" :)

Just one defensive suggestion.

texastony (changes requested):
I have several questions and one blocking request.

texastony:
@ShubhamChaturvedi7 This may unblock your use of https://github.com/aws/aws-cryptographic-material-providers-library/compare/main...tony/feat-update-ddb-model

PR #257, opened 2023-05-16
feat: Explicitly reject unsupported shapes, traits, etc.

### LINKED ISSUE #185: Explicitly reject unsupported shapes/traits
E.g. `@streaming` is currently silently ignored, leading to type errors (e.g. `System.IO.Stream` vs `System.IO.MemoryStream`).

~Part of a more general class of "silently does the wrong thing on unrecognized shapes/traits" bugs - is there a more generic way code generation could fail closed on unsupported things?~

I suspect we can write a selector expression for all shapes and traits this generator supports, and just do an initial application of this selector to every model and fail if anything is not covered. This would be similar to the selector the codegen README suggests using to filter out such unsupported features.

robin-aws:
*Issue #, if available:* 

#185

*Description of changes:*

Adds a model validator with allow-lists for shapes and traits the code generators understand, meaning that we are aware of the meaning of the shape or trait and believe that the correct code generation is implemented. For many service traits this just means we've confirmed that they don't change the API callers use when in `--aws-sdk` style mode, but some such as `@streaming` definitely do affect this API and hence are not in the list.

We could have specific lists for individual target languages as well in the future, but this will be more difficult to fit into the validation framework, which isn't aware of which target language we're building for. Dividing up the codegen into separate plugins with their own validators would be the natural way to address this.

I populated these lists by adding all necessary entries to cover the TestModels, but also tested on all downstream projects via `make polymorph_dafny`, which did reveal a couple of tweaks not covered by the test models.

Resolves #185


texastony (dismissed):
There are bold statements here about what we support.

texastony (commented):
I approve of the README changes.
But there are still issues with the supported traits.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
```diff
@@ -374,4 +388,127 @@ public enum TargetLanguage {
...
+                        ")";
+
+        Set<Shape> notSupported = Selector.parse(selectorExpr).select(model);
+
+        if (!notSupported.isEmpty()) {
+            String message = "The following shapes in the service's closure are not supported: \n" +
+                    notSupported.stream()
+                                .map(shape -> unsupportedShapeLine(shape, supportedShapes, supportedTraits))
+                                .collect(Collectors.joining("\n"));
+            throw new IllegalArgumentException(message);
```
robin-aws: Note to self: it will be much nicer to make these `DANGER`-level Validator events instead. That means we'll get an event for every unsupported thing, and the ability to suppress them if a user checks that the trait is not relevant or the lack of support is harmless.

There is even precedent for individual code generators providing validators to check for their own limitations:
* Rust: https://github.com/smithy-lang/smithy-rs/blob/14ccc50f41ea8b59199fecf239131826770fa0d1/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/PatternTraitEscapedSpecialCharsValidator.kt#L16
* Typescript: https://github.com/smithy-lang/smithy-typescript/blob/3c9c07ba1784e1b0a04bb44a09890335378ea21b/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/validation/LongValidator.java#L38

robin-aws: Note that some things we don't support should be hard `ERROR` events, which are for fatal errors that prevent processing a model and can't be suppressed. We don't support Smithy 2.0 `enum` shapes in .NET for example, and that can't be worked around.

seebees (dismissed):
LGTM

seebees (dismissed):
LGTM

seebees (approved):
Still looks good :)

PR #277, opened 2023-07-05
fix(NET): NET SDK Error Wrapping

josecorella:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/TypeConversionCodegen.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/TypeConversionCodegen.java
@@ -6,14 +6,7 @@ package software.amazon.polymorph.smithydotnet;
 import com.google.common.annotations.VisibleForTesting;
 
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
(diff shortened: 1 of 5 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/TypeConversionCodegen.java
```diff
@@ -461,6 +454,26 @@ private String generateConstructorArg(final MemberShape memberShape) {
...
+                "InvalidAlgorithmSuiteInfo",
+                "InvalidAlgorithmSuiteInfoOnDecrypt",
+                "InvalidAlgorithmSuiteInfoOnEncrypt",
+                "InvalidDecryptionMaterials",
+                "InvalidDecryptionMaterialsTransition",
+                "InvalidEncryptionMaterials",
+                "InvalidEncryptionMaterialsTransition",
+                "KeyStoreException",
+                "KeyVectorException"
+        );
```
texastony: Style: Move this to a constant somewhere. 
Ideally, this list would be a parameter of a method in the call chain,
but that can be a refactor for an other day.

Smithy-Dafny is trying to support users other than Crypto Tools.

We MUST NOT casually hard code custom logic.

texastony (changes requested):

texastony (approved):

PR #282, opened 2023-08-15
Python POC

lucasmcdonald3:
Python codegen for Smithy-Dafny.

This generates conversion code between Dafny-generated Python code and

1) Polymorph LocalServices
2) Polymorph wrapped LocalServices
3) AWS SDKs (boto3 wrapper)

The generated code works for all TestModels, the MPL, and MPL TestVectors.
MPL PR: https://github.com/aws/aws-cryptographic-material-providers-library/pull/171

## Sub-PRs

These smaller PRs focus on parts of this PR. These are not functional on their own. These are solely intended for reviewer readability.

* Smithy/Dafny conversion. This contains ShapeVisitors that generate code to convert between Smithy and Dafny shapes and NameResolver helpers. https://github.com/lucasmcdonald3/smithy-dafny/pull/12/
* TestModel project structure. This contains a file structure layout for TestModels  and Python SharedMakefile targets. We would use a similar layout for all Polymorph projects. (This is pending author update -- I plan to change 
namespacing soon.) https://github.com/lucasmcdonald3/smithy-dafny/pull/9
* Customize code. This contains code that is called from the `customize` directive. This writes new files, and generally does not need to override any Smithy-Python generated code. https://github.com/lucasmcdonald3/smithy-dafny/pull/10. 
* Polymorph CLI integration. This allows Python codegen to be used as a target of Polymorph CLI. https://github.com/lucasmcdonald3/smithy-dafny/pull/13
* Overwriting and extending Smithy-Python code. This contains code that overwrites Smithy-Python's behavior to make an HTTP request. It also overwrites its request serialization (Smithy to Dafny) and deserialization (Dafny to Smithy) logic. https://github.com/lucasmcdonald3/smithy-dafny/pull/11
  * Related: Changes to forked Smithy-Python. This contains changes to Smithy-Python, mostly to make it more extensible for our use case. https://github.com/lucasmcdonald3/smithy-python/pull/1

## Pending work

* "ExternV2"
  * Replace PYTHONPATH hacks with Dafny ModuleName. These are identified with `TODO-Python-PYTHONPATH`. 
  * Change generated extern names back from `_` to `.` (this will fix Java and Net AWS SDK GHA failures)
* Use PyPi DafnyRuntime
* Timestamp shapes (not used in MPL I believe; will be needed for DBESDK)

## Build instructions 

Set up project:
* Check out fork with submodules
* Build Dafny-nightly-latest from source; set as your default Dafny
* Install Python 3.11

Build a TestModel:
```
make polymorph_dafny
make polymorph_python
make transpile_python
make test_python
``` 

## GHA Failures
* **NET/Java/Verification AWS SDKs:** I committed a name change to AWS SDKs to help Python out. I will remove this as part of Dafny ModuleName work ("ExternV2").
* **smithy-dafny-codegen-cli:** This should solve itself once Smithy-Python is on Maven central, so I'm not planning to debug it (unless this becomes blocking).

## Misc

* This PR is WIP. Its primarily provides a testbed for me during development.
* I won't merge this in until all CI passes to avoid blocking mainline Smithy-Dafny on in-dev work.
* Smithy-Python is included as a submodule. It points to [my fork of Smithy-Python](https://github.com/lucasmcdonald3/smithy-python/tree/smithy-dafny-mainline). I am not sure how to proceed with having this fork of Smithy-Python. Ideally, we would include Smithy-Python's mainline as a submodule. I will clean up the changes on my fork, get those changes looked at by us, then open a PR against Smithy-Python. 


robin-aws (changes requested):
Sticking my foot in the door as I really want to review this - feel free to pester me if I haven't started in a week :)

.github/workflows/smithy-dafny-conversion.yml
```diff
@@ -11,10 +11,13 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
       - uses: actions/setup-java@v3
         with:
           distribution: 'corretto'
-          java-version: '8'
+          java-version: '17'
```
texastony: Issue/Blocking: Revert this.

The Conversion Library is a runtime dependency of products that MUST support Java 8.
As such, it is built and tested with Java 8.

The code generators are not runtime dependencies,
and use the latest and greatest Java features.

TestModels/Aggregate/runtimes/python/test/extern/test_dafny_wrapper.py
texastony, line 1: Praise: 
I think this is very neat, readable code. 

Question:
I wonder if we should move the `test_dafny` method in to a helper python module,
that all the TestModels could then load. 
If we ever need to adjust it,
we would only have to modify it once.

TestModels/Aggregate/smithy-build.json
texastony, line 1: Question:
Does Smithy-Dafny-Python use this?
If so, Python is ahead of Java and .NET.

TestModels/Constraints/runtimes/python/pyproject.toml
texastony, line 1: Note: I am skipping over the rest of the test Models for now.

TestModels/SharedMakefile.mk
```diff
@@ -297,5 +351,94 @@ _clean:
...
+# This is tracked in https://github.com/dafny-lang/dafny/issues/4322.
+# This may require new Smithy-Dafny logic to generate Pythonic extern names.
+_python_underscore_extern_names:
+	find src -regex ".*\.dfy" -type f -exec sed -i $(SED_PARAMETER) '/.*{:extern \".*\".*/s/\./_/g' {} \;
+	find Model -regex ".*\.dfy" -type f -exec sed -i $(SED_PARAMETER) '/.*{:extern \".*\.*"/s/\./_/g' {} \;
+	find test -regex ".*\.dfy" -type f -exec sed -i $(SED_PARAMETER) '/.*{:extern \".*\".*/s/\./_/g' {} \;
+
+_python_underscore_dependency_extern_names:
+	$(MAKE) -C $(STANDARD_LIBRARY_PATH) _python_underscore_extern_names
+	$(patsubst %, $(MAKE) -C $(PROJECT_ROOT)/% _python_underscore_extern_names;, $(LIBRARIES))
```
texastony: Praise/Note: 
Nice way to "Make it Work". 
This is very close to solution written by an ARG Intern from 2019.

texastony (commented):
Praise: I have only viewed 15 files, but I like what I have read so far.

TestModels/Aggregate/runtimes/python/test/extern/test_dafny_wrapper.py
lucasmcdonald3: Will do. I hope to remove the `test_dafny` entirely. I wrote this before I figured out how to support externs, and I think I don't need this anymore if I set up externs correctly.

TestModels/Aggregate/smithy-build.json
lucasmcdonald3: Yes

TestModels/Aggregate/runtimes/python/test/extern/test_dafny_wrapper.py
lucasmcdonald3: Updated: This is much simpler now.

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
```diff
@@ -66,13 +80,54 @@ public static void main(String[] args) {
...
+                objectNodeBuilder.withMember(entry.getKey(), entry.getValue().toString().replace("\"", ""));
+            }
+        }
+
+        final ServiceShape serviceShape = ModelUtils.serviceFromNamespace(serviceModel, cliArguments.namespace);
+        final PluginContext pluginContext = PluginContext.builder()
+            .model(serviceModel)
+            .fileManifest(FileManifest.create(cliArguments.outputPythonDir().orElse(cliArguments.modelPath)))
+            .settings(objectNodeBuilder.build())
+            .build();
```
texastony: Question/Issue: 
I thought Robin and Alex had already implemented
a `smithy-build.json` parser.

It is probably not in the CLI project,
but the actual codegen project.

See: https://github.com/smithy-lang/smithy-dafny/tree/main-1.x/TestModels/aws-sdks/sqs

Also, here is where that Test Model is executed in CI:
https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/.github/workflows/smithy-polymorph.yml#L31-L42

lucasmcdonald3: I typed out a longer response, but I lost that somewhere in my tabs...
I believe they did not write a parser, but are instead using the standard Smithy `gradle` build process, which parses it for them.
We still need a way to pipe this information into the CLI.

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/standard_library/extern/UTF8.py
```diff
@@ -0,0 +1,86 @@
...
+
+'''
+Extern UTF8 encode and decode methods.
+
+Note:
+Python's `.encode('utf-8')` does not handle surrogates.
+These methods are expected to handle surrogates (e.g. "\uD808\uDC00").
+To work around this, we encode Dafny strings into UTF-16-LE (little endian)
+and decode them before encoding into UTF-8 (`_strict_tostring`).
+To decode, we reverse the encode step. (`_reverse_strict_tostring`)
```
texastony: Question: Tell me about the performance characteristics of this implementation.
It appears that we parse every character three times.

For Encoding:
1. Encode to UTF-16-LE
2. Decode to ?
3. Encode to UTF-8

For Decode:
1. Decode from UTF-8

Is this correct?

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/standard_library/extern/UTF8.py
```diff
@@ -0,0 +1,86 @@
...
+    # Catch both Encode and Decode errors.
+    # The `try` block involves both encoding and decoding.
+    except (UnicodeEncodeError, UnicodeDecodeError):
+      return Wrappers.Result_Failure(_dafny.Seq("Could not encode input to Dafny Bytes."))
+
+  @staticmethod
+  def _strict_tostring(s):
+    '''
+    Converts a Dafny Seq of ASCII characters
+    into a string by writing a UTF-16-LE decoding,
```
texastony: From above, I assume this is a typo:
```suggestion
    into a string by writing a UTF-16-LE encoding,
```

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/standard_library/extern/UTF8.py
```diff
@@ -0,0 +1,86 @@
...
+    then decoding it to ASCII with `strict` encoding.
+
+    This encoding-decoding allows subsequent ASCII-to-UTF8 encodings
+    to handle surrogates as expected by Dafny code.
+
+    This is exactly the `_dafny.string_from_utf_16` method, except with
```
texastony: Sorry, I am new to Dafny Python.
Where can I read this `_dafny.string_from_utf_16` method?

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/standard_library/extern/UTF8.py
```diff
@@ -0,0 +1,86 @@
...
+    for i in range(int(len(b)/2)):
+      # Take two consecutive bytes;
+      c = b[2*i:2*i+2]
+      # Unpack them into an ordinal representation;
+      d = struct.unpack('<H', c)
+      # Convert into a character representation;
+      e = chr(d[0])
+      # Append to returned string
+      out.append(e)
+    return _dafny.Seq(out)
```
texastony: There is enough logic here that I would like some more testing.
I committed a couple of UTF-8 text files to the conversion library:
https://github.com/smithy-lang/smithy-dafny/tree/main-1.x/smithy-dafny-conversion/src/test/resources

We could re-use those to ensure that we have this right.

Tests: https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/smithy-dafny-conversion/src/test/java/software/amazon/smithy/dafny/conversion/ToDafnyTest.java#L122-L142

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
texastony: Oh, I would argue that we just should not.
Either we use the CLI without the JSON,
or we support the JSON via the gradle build process.

Otherwise, it feels like we are not adopting Smithy conventions but committing a work around to them...

But we can bring this up in a meeting.

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/standard_library/extern/UTF8.py
lucasmcdonald3: Here: https://github.com/dafny-lang/dafny/blob/988b41cf36593e4fa7faeebfa839f243b541e043/Source/DafnyRuntime/DafnyRuntime.py#L22

(This is the DafnyRuntime file for Python.)

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/standard_library/extern/UTF8.py
lucasmcdonald3: We are parsing each character 3 times on decode as well (reverse of encode).

(Obviously I'm not happy about this, but I struggled to handle surrogate characters, and this seemed to be the most commonly suggested pattern to handle those...)

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/standard_library/extern/UTF8.py
lucasmcdonald3: I did some testing here. I added both files to my this `test/extern` folder and wrote a new `test_utf8_files.py` script to test them.

* UTF-8-Demo: Pass
* UTF-8-test: I think this is expected to fail, but Python's `open` cannot open this file. It fails at the line
```
2.1.5  5 bytes (U-00200000):        "�����"
```
I ended up copy-pasting this into the file so I could debug encoding/decoding failure modes and added better exception handling.

Test code: [extern.zip](https://github.com/smithy-lang/smithy-dafny/files/12795611/extern.zip)

texastony: I should dig into why Java just passes UTF-8-test.

texastony:
Note: https://github.com/smithy-lang/smithy-dafny/blob/lucmcdon/python-poc/.github/workflows/test_models_dafny_verification.yml#L87

It looks like the Python GHA uses Dafny 4.3 but the Dafny Verification uses 4.1.

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/standard_library/extern/UTF8.py
texastony: We do not need to block development on this.
But we should not ship a product with this deficiency.

I do not know how often we convert strings,
but I imagine it's often.

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
robin-aws: The CLI and the Smithy plugin both pipe their configuration through the common `CodegenEngine` class, so to support a new language you should only have to augment the set of CLI options and tweak the `DafnyClientCodegenPluginSettings` class (which is the deserialized model of the relevant sub-section of `smithy-build.json` - see `DafnyClientCodegenPluginSettings.fromObject`).

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/standard_library/extern/UTF8.py
robin-aws: This might be a good time to at least discuss the eventual path to using `--unicode-char:true` in Dafny. A big part of supporting that mode is avoiding exactly this annoying edge case.

robin-aws:
> Smithy-Python is included as a submodule. It points to [my fork of Smithy-Python](https://github.com/lucasmcdonald3/smithy-python/tree/smithy-dafny-mainline). I am not sure how to proceed with having this fork of Smithy-Python. Ideally, we would include Smithy-Python's mainline as a submodule. I will clean up the changes on my fork, get those changes looked at by us, then open a PR against Smithy-Python.

What changes did you have to make to `smithy-python`? If we can get those contributed upstream or find a way to not need them, you could just depend on the `software.amazon.smithy.python:smithy-python-codegen` package from Maven Central (I'm not sure when they are planning on publishing it there but it's likely pretty soon).

lucasmcdonald3:
> What changes did you have to make to smithy-python? If we can get those contributed upstream or find a way to not need them, you could just depend on the software.amazon.smithy.python:smithy-python-codegen package from Maven Central (I'm not sure when they are planning on publishing it there but it's likely pretty soon).

Changes are [here](https://github.com/lucasmcdonald3/smithy-python/pull/1).

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
lucasmcdonald3: This is updated. Check out https://github.com/lucasmcdonald3/smithy-dafny/pull/13

PR #300, opened 2023-10-17
feat: Repair OpaqueError

ajewellamz:
*Issue #, if available:*

*Description of changes:*

OpaqueError has two constructors :
      OpaqueError(Exception ex) : base("OpaqueError:", ex) { this.obj = ex; }
      OpaqueError(object obj) : base("Opaque obj is not an Exception.") { this.obj = obj;}

Unfortunately, overloading is resolved at compile time, so if you have an Exception currently cast to an Object, you get the second one, and therefore lose all the information in the original Exception.
Even more unfortunately, the thing that turns an extern's Error_Opaque into an OpaqueError does exactly that : casts the original Exception to an Object before use.

The solution is to change the second constructor to behave differently if the Object is really an Exception

OpaqueError(object obj)
  : base(obj is Exception ? "OpaqueError:" : "Opaque obj is not an Exception.", obj as Exception)
{ this.obj = obj;}

"obj as Exception" is either an Exception, in which case we get back the context, or null, which is exactly the old behavior.


texastony (approved):
Great PR description!
Really helps clarify what is going on

PR #301, opened 2023-10-20
feat: add type descriptors for data constructors in Java, when necessary

robin-aws:
*Description of changes:*
As part of the progress towards supporting traits on non-reference types, the Java backend in Dafny 4.2 changed datatype constructors to require type descriptors for any generic type parameters: https://github.com/dafny-lang/dafny/pull/4240.

This change fixes a big chunk of the failing nightly build failures by augmenting the Java code generation to conditionally provide these type descriptors when creating `Result` and `Option` values. In most cases this is handled by the new `nameresolver.Dafny.createSuccess/Failure/Some/None` helper methods, but in some cases some new Dafny helper methods are added to `StandardLibrary` or emitted as well, so that TestModels can provide the same Java code for any Dafny version.

It also adds a parameter for the target Dafny version so it knows when to emit them, exposed both as a `--dafny-version` option for the CLI and a `dafnyVersion` configuration value on `smithy-build.json` file. This defaults to `4.1.0`, but the latter knob is actually required when using the latest "edition" of the Smithy plugin (now `2023.10` instead of `2023`), as this is exactly the kind of otherwise-breaking change editions are designed to support!

Finally, it refactors the CI to test all Dafny-relevant things on multiple versions of Dafny. It follows the same workflow template as https://github.com/aws/aws-cryptographic-material-providers-library-java/tree/main/.github/workflows (and kudos to @texastony!). This exposed one verification regression on 4.3 that I'm looking into, and may factor out of this PR if it's not quick to fix. ~Note the nightly build depends on a pending patch to `setup-dafny-action` to provide the actual version when using `nightly-latest`: https://github.com/dafny-lang/setup-dafny-action/pull/19~


.github/workflows/pull.yml
```diff
@@ -0,0 +1,37 @@
...
+on:
+  pull_request:
+
+jobs:
+  pr-ci-verification:
+    strategy:
+      matrix:
+        dafny-version: [
+          4.1.0,
+          4.3.0
```
seebees: You talk about 4.2 in the PR but don't test it here?

Also, given the repetition of version you may want to pull that out as well.
Just saying

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
```diff
@@ -119,6 +120,11 @@ private static Options getCliOptions() {
             .desc("<optional> AWS SDK for Java version to use: v1, or v2 (default)")
             .hasArg()
             .build())
+          .addOption(Option.builder()
+            .longOpt("dafny-version")
+            .desc("<optional> Dafny version to use. Defaults to 4.1.0")
```
seebees: You sure about that? I feel like defaulting to 4.1 is something we may regret

seebees (commented):
How is all this going to play with a fully featured conversion library?
Are we going to come back and refactor externs again?

.github/workflows/pull.yml
robin-aws: Fair. Although getting to 4.3 across the board is more important than 4.2 in the broader context.

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
robin-aws: I was only trying to be backwards compatible for anyone else using the CLI. Note that the smithy plugin avoids this by incrementing the latest edition and making it required!

robin-aws:
> How is all this going to play with a fully featured conversion library?
> Are we going to come back and refactor externs again?

In this particular case it's a nasty thing to try and put in the conversion library, because this is exactly the information that Java loses at runtime: you can't provide a Java `createSome<T>(T value)` method because it can't in general figure out what `T` was at compile time precisely enough to create the Dafny `TypeDescriptor` for it.

You can see in some places I've defined specific, non-generic Dafny functions like `createStringSome(s: string): Option<string>` because then Dafny itself provides the type descriptors for you. That's actually the right long-term approach to this: only define the conversion logic in Dafny instead of target language code, so you're no longer coupled to these Dafny internals.

.github/workflows/pull.yml
robin-aws: I was actually a little worried about creating so many individual GHA jobs by testing on more than 2. As part of a different PR I'd like to first get the parameterization of test models out of workflow files: https://github.com/smithy-lang/smithy-dafny/pull/257/files#diff-a933719600e0ce4e5ada8a6a7a64dfb38e005f6145c00147d9a02a16c69409d7R20-R24 After that I'd be happier to test on more dafnies.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/Shim.java
```diff
@@ -0,0 +1,117 @@
...
+import com.squareup.javapoet.TypeName;
+import software.amazon.polymorph.smithyjava.generator.CodegenSubject;
+import software.amazon.polymorph.smithyjava.generator.Generator;
+import software.amazon.polymorph.smithyjava.nameresolver.Dafny;
+import software.amazon.smithy.model.shapes.ShapeType;
+
+import static javax.lang.model.element.Modifier.PUBLIC;
+import static javax.lang.model.element.Modifier.STATIC;
+
+public abstract class Shim extends Generator {
```
texastony: Java Docs or other comments are required.
Polymorph is a wild, complicated, library.
Give people some idea of:
- when these methods SHOULD called
- what methods SHOULD be in this class

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v1/ShimV1.java
```diff
@@ -18,24 +19,22 @@
...
 import software.amazon.smithy.model.shapes.OperationShape;
 import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.ShapeId;
 import software.amazon.smithy.utils.StringUtils;
 
-import static software.amazon.polymorph.smithyjava.nameresolver.Constants.DAFNY_RESULT_CLASS_NAME;
 import static software.amazon.polymorph.smithyjava.nameresolver.Constants.DAFNY_TUPLE0_CLASS_NAME;
 import static software.amazon.polymorph.smithyjava.nameresolver.Constants.SMITHY_API_UNIT;
 
-//TODO: Create abstract class for V1 & V2 to extend
```
texastony: Restore this TODO as:
"compete V1 & V2 abstraction"

Yes, they now have an abstract class.
But they still have a ton of duplicate logic.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v1/ShimV1.java
texastony, line 1: FYI: V1 is not tested in any way.
If you wanted, you could delete the whole thing.
I thought we were going to try and support Keyrings with the AWS SDK V1.
But we are not.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/ShimV2.java
```diff
@@ -18,24 +19,22 @@
...
 import software.amazon.smithy.model.shapes.OperationShape;
 import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.ShapeId;
 import software.amazon.smithy.utils.StringUtils;
 
-import static software.amazon.polymorph.smithyjava.nameresolver.Constants.DAFNY_RESULT_CLASS_NAME;
 import static software.amazon.polymorph.smithyjava.nameresolver.Constants.DAFNY_TUPLE0_CLASS_NAME;
 import static software.amazon.polymorph.smithyjava.nameresolver.Constants.SMITHY_API_UNIT;
 
-//TODO: Create abstract class for V1 & V2 to extend
```
texastony: See other comments about incomplete abstraction and possible just removing V1.
Though, even if we removed V1, we may want to leave an abstract class.
Who knows when V3 will come around.

texastony (changes requested):

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v1/ShimV1.java
robin-aws: Ah I was wondering, thanks, if they are not being used anywhere yes I'd rather delete them entirely.

codegen/smithy-dafny-codegen/src/test/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/Constants.java
```diff
@@ -34,43 +58,167 @@ public Result<Tuple0, Error> DoVoid(DoVoidRequest input) {
...
+                    return Option.create_None(DafnySequence._typeDescriptor(TypeDescriptor.CHAR));
+                  }
+                          
+                  public static Option<Boolean> createBooleanSome(Boolean b) {
+                    return Option.create_Some(TypeDescriptor.BOOLEAN, b);
+                  }
+                              
+                  public static Option<Boolean> createBooleanNone() {
+                    return Option.create_None(TypeDescriptor.BOOLEAN);
+                  }
```
texastony: I do not understand what these are doing here.
If they are static with respect to the Dafny Version,
they should be in the conversion library.

We can publish a conversion library for 4.2 that contains all of these.

We MUST NOT generate a  bunch of methods over and over again.

codegen/smithy-dafny-codegen/src/test/java/software/amazon/polymorph/util/Tokenizer.java
```diff
@@ -33,6 +34,11 @@ public static record ParseToken(String text, int type) {}
     public static void tokenizeAndAssertEqual(String expected, String actual) {
         final List<ParseToken> actualTokens = tokenize(actual);
         final List<ParseToken> expectedTokens = tokenize(expected);
-        Assert.assertEquals(expectedTokens, actualTokens);
+        if (!Objects.equals(expectedTokens, actualTokens)) {
+            // If the tokens aren't equal, assert the original strings are equal
+            // knowing it is guaranteed to fail,
+            // just so that we get a much more readable diff in tooling.
+            Assert.assertEquals(expected, actual);
+        }
```
texastony, line 42: Massive Praise: I really wish I had thought of this.
Would have made my life so much easier.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/ShimV2.java
```diff
@@ -71,6 +70,12 @@ TypeSpec shim() {
                 .addField(
                         ClassName.get(String.class),
                         "region", Modifier.PRIVATE, Modifier.FINAL)
+                .addMethod(successOfClientConstructor())
+                .addMethod(failureOfErrorConstructor())
+                .addMethod(stringSomeConstructor())
+                .addMethod(stringNoneConstructor())
+                .addMethod(booleanSomeConstructor())
+                .addMethod(booleanNoneConstructor())
```
texastony: I have been trying to figure out why we need these.
If we need them for AWS SDK, why don't we need them for other Shims?

As far as I can tell, we do not need them at all...
Alternatively, if we always need them, they should be in the conversion library.

texastony (dismissed):

texastony:
@robin-aws I put this PR back in draft.
Once you are ready for another review, mark it as ready and ping the channel.

codegen/smithy-dafny-codegen/src/test/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/Constants.java
robin-aws: Okay I see now in the documentation for the conversion library is that the intent was to update it for new Dafny versions. Now that I got all this working I'll move these into a new version of it.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/ShimV2.java
robin-aws: It's only necessary because of the values of those types created in code like this: https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/TestModels/aws-sdks/kms/runtimes/java/src/main/java/software/amazon/cryptography/services/kms/internal/__default.java, AFAICT the other shims don't need to handle failures to create clients or the region match feature.

As above I'll add them to the conversion library instead though.

texastony: But that file is entirely manually written.
It is not generated by Smithy-Dafny.
So generating these methods here will 
only effect that code IF
we manually update it to use them.

Finally, while we only have problem RIGHT NOW in these classes,
(which I believe is an assumption)
we may have the same problem else where SOON.

As such, I'd favor a 4.3 only fix.

Smithy-Dafny can throw "generation" time Error for 4.2.
We only need to publish a new Conversion library for 4.3.

Not putting these into the conversion library limits
the utility of this fix.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/Shim.java
robin-aws: Ended up not needing this class at all in the new approach

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v1/ShimV1.java
robin-aws: Since I didn't end up touching the shim generators at all, I elected not to nuke V1 just yet to keep this PR smaller.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v1/ShimV1.java
robin-aws: Put the TODO back as is since I reverted most of these changes

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/ShimV2.java
robin-aws: As per offline conversation, it turns out to be not possible to add these methods to the conversion library, since `Option` and `Result` are not part of the Dafny runtimes, so the conversion library can't reference them without strange dependency machinations.

However, the better solution turns out to be to define them in Dafny source code instead, either in `StandardLibrary` (for the boolean and string Options) or the Dafny abstract modules (for the `Result<ITheServiceClient>` types).

codegen/smithy-dafny-codegen/src/test/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/Constants.java
robin-aws: See https://github.com/smithy-lang/smithy-dafny/pull/301/files#r1380740881

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
robin-aws: After offline conversation, elected to make it required in the CLI as well.

TestModels/dafny-dependencies/StandardLibrary/src/WrappersInterop.dfy
```diff
@@ -0,0 +1,34 @@
...
+// in the Java backend across different versions of Dafny,
+// but may be useful for other target languages in the future as well.
+//
+module StandardLibrary.Interop {
+
+  import opened Wrappers
+
+  class WrappersInterop {
+
+    static function method CreateStringSome(s: string): Option<string> {
```
seebees, line 32: Generics? `CreateSome<T>(t: T): Option<T>`?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
```diff
@@ -2026,23 +2028,57 @@ public TokenTree generateAbstractAwsServiceClass(ServiceShape serviceShape) {
...
-            "returns (res: Result<%s, Error>)".formatted(nameResolver.traitForServiceClient(serviceShape)),
+            "returns (res: Result<%s, Error>)".formatted(dafnyClientTrait),
             "ensures res.Success? ==> ",
             "&& fresh(res.value)",
             "&& fresh(res.value.%s)".formatted(nameResolver.mutableStateFunctionName()),
             "&& fresh(res.value.%s)".formatted(nameResolver.callHistoryFieldName()),
             "&& res.value.%s()".formatted(nameResolver.validStateInvariantName())
           ).lineSeparated();
 
+        final TokenTree createSuccessOfClient = TokenTree
```
seebees: I'm a little confused about where this is getting emitted.
On the one hand, I see additions in the "StandardLibrary" but I also see this getting written here?

TestModels/dafny-dependencies/StandardLibrary/src/WrappersInterop.dfy
robin-aws: Unfortunately no, but I beefed up the module comment to explain why better.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
robin-aws: Similarly added more comments to explain better.

.github/workflows/pull.yml
texastony: Note: I am still learning how to best utilize GHA/Ws,
and it seems that GitHub is evolving them as well.

And our test models need to be refactored to an integration test project...

.github/workflows/pull.yml
```diff
@@ -0,0 +1,40 @@
...
+    with:
+      dafny: ${{ matrix.dafny-version }}
+  pr-ci-java:
+    strategy:
+      fail-fast: false
+      matrix:
+        dafny-version: [
+          4.1.0,
+          4.3.0
+        ]
```
texastony: Nit: Is there no way to declare this once and reference it elsewhere in this file?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
```diff
@@ -1883,8 +1884,7 @@ public TokenTree generateAbstractLocalService(ServiceShape serviceShape)  {
                 ),
             // Yes, Error is hard coded
             // this can work because we need to be able Errors from other modules...
-            "returns (res: Result<%sClient, Error>)\n"
-                .formatted(localServiceTrait.getSdkId())
+            "returns (res: Result<%s, Error>)\n".formatted(dafnyClientTrait)
```
texastony, line 1887: Issue: I am a little alarmed about the change from `sdkId` to trait.
Is it always the case these are equal/the same?
Or was this code wrong to ever use `getSdkId`?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyVersion.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.Objects;
+
+/**
+ * Representation of a Dafny version number, according to SemVer 1.0 semantics.
+ *
+ * Note that Dafny pre-releases historically have used pre-release suffixes slightly wrong:
+ * after releasing 4.2 for example, the nightly pre-releases will have version numbers like
+ * "4.2.0-nightly-2023-08-04-656a114", but that should actually be interpreted as a pre-release
+ * for 4.2 rather than 4.3. So far that's immaterial for this code base,
+ * but if it becomes relevant the better solution is for Dafny pre-releases to correct this instead.
```
texastony, line 15: Nit: This note is confusing.
Did you mean: 

> but that should actually be interpreted as a pre-release for **4.3** rather than **4.2**.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyVersion.java
texastony, line 1: This looks right, is there a unit test? YES.
Most Groovy.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/library/shims/ServiceShim.java
```diff
@@ -125,7 +125,7 @@ private MethodSpec serviceConstructor(BuilderSpecs builderSpecs) {
         method.addStatement(dafnyDeclareAndConvert(trait.getConfigId()));
         // Invoke client creation
         // i.e.: Result<AtomicPrimitivesClient, Error> result = software.amazon.cryptography.primitives.internaldafny.__default.AtomicPrimitives(dafnyValue);
-        TypeName success = subject.dafnyNameResolver.classNameForConcreteServiceClient(subject.serviceShape);
+        TypeName success = subject.dafnyNameResolver.classNameForInterface(subject.serviceShape);
```
texastony, line 128: Question/Issue: This is a large change, as the result type is now the abstract interface,
as compared to the concrete instance.

~~If we defined any methods on the concrete that are not included on the abstract,
we will be loosing access to them,
with out a cast.~~
~~Do you know why this change was necessary?~~

Having read the rest of the file, 
I see that we store it as an interface anyway.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/library/shims/TestServiceShim.java
texastony, line 1: Why were imports added but not used?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/Constants.java
```diff
@@ -11,6 +11,7 @@
 
 public class Constants {
     public static final ShapeId SMITHY_API_UNIT = ShapeId.fromParts("smithy.api", "Unit");
+    public static final ClassName DAFNY_OPTION_CLASS_NAME = ClassName.get("Wrappers_Compile", "Option");
```
texastony, line 14: Not I have seen the been used yet.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/Dafny.java
```diff
@@ -92,6 +97,74 @@ public static String datatypeConstructorCreate(String name, boolean isRecordType
         return "create_" + DafnyNameResolverHelpers.dafnyCompilesExtra_(name);
     }
 
+    private static final DafnyVersion NEEDS_TYPE_DESCRIPTORS_WHEN_CONSTRUCTING_DATATYPES = new DafnyVersion(4, 2, 0);
```
texastony, line 101: Nit: Shouldn't this be a collection? 
Such that multiple Dafny versions can fit this bill?

Update: OH, Dafnys greater than or equal to this will need it.

```suggestion
    //  Dafnys greater than or equal to this will need Type Descriptors for constructing datatypes  
    private static final DafnyVersion NEEDS_TYPE_DESCRIPTORS_WHEN_CONSTRUCTING_DATATYPES = new DafnyVersion(4, 2, 0);
```

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/Dafny.java
```diff
@@ -92,6 +97,74 @@ public static String datatypeConstructorCreate(String name, boolean isRecordType
...
+                    typeDescriptor,
+                    error);
+        } else {
+            return CodeBlock.of(
+                    "$T.create_Failure($L)",
+                    Constants.DAFNY_RESULT_CLASS_NAME,
+                    error);
+        }
+    }
+
```
texastony: Blocking: Add Java docs.
Smithy-Dafny is a wildly confusing project with many hands contributing functionality.
We MUST describe what our methods do.

I checked, its true, all of these method names are unique.
But they are barely unique.

We need to java docs so future us know when to invoke this these methods.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/Dafny.java
```diff
@@ -259,6 +339,19 @@ public CodeBlock typeDescriptor(ShapeId shapeId) {
...
+        }
+        if (shape.getType().isShapeType(ShapeType.SET)) {
+            CodeBlock elementTypeDescriptor = typeDescriptor(shape.asSetShape().get().getMember().getTarget());
+            return CodeBlock.of("$T._typeDescriptor($L)", Constants.DAFNY_SET_CLASS_NAME, elementTypeDescriptor);
+        }
+        if (shape.getType().isShapeType(ShapeType.MAP)) {
+            CodeBlock keyTypeDescriptor = typeDescriptor(shape.asMapShape().get().getKey().getTarget());
+            CodeBlock valueTypeDescriptor = typeDescriptor(shape.asMapShape().get().getValue().getTarget());
+            return CodeBlock.of("$T._typeDescriptor($L, $L)", Constants.DAFNY_MAP_CLASS_NAME, keyTypeDescriptor, valueTypeDescriptor);
+        }
```
texastony, line 378: Question: Where is the  Dafny 4.2.0 switch?

This is a public method.
So it can be called by another class that is generating for less than 4.2.0.

Is this logic invalid for Dafny less than 4.2.0?
If so, we need the check.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/Dafny.java
```diff
@@ -304,7 +397,7 @@ TypeName typeForString(StringShape shape) {
         return typeForCharacterSequence();
     }
 
-    TypeName typeForCharacterSequence() {
+    public TypeName typeForCharacterSequence() {
```
texastony: Note for Tony: I need to pull this down and grep around to see if this needs to be public.

texastony (changes requested):
Java docs on new public methods is the only blocking comment.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/library/shims/ServiceShim.java
texastony: I saw that we had to update to the `index.dfy` files to return the interface.
This may bite us later, but we will resolve the issue then.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/Dafny.java
texastony: I pulled this PR down and found no public uses of this.
Lets leave this as private as possible.

```suggestion
   TypeName typeForCharacterSequence() {
```

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/Dafny.java
robin-aws: This method just emits code to create type descriptor instances, and the way those are created didn't change, only where they are required parameters. So this logic is correct for any Dafny version.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/Dafny.java
robin-aws: Good call, 'twas just leftover from earlier approach

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
robin-aws: Ah the real change is changing from the concrete client class type (e.g. `SimpleLocalServiceClient`) to the trait type instead (e.g. `ISimpleLocalServiceClient`). That helps use the same helper methods for the wrapped client case as well, and seems like a good encapsulation pattern anyway.

texastony: God... what bad naming on our part...
I guess it makes sense since a Dafny Abastract Class is Called a Trait.
But I thought it was refering to the Smithy Trait.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/Dafny.java
texastony: But we have separate logic for injecting Type Descriptors in to Aggregate Smithy Types, 
right?
Does this get called on those paths?
I guess so, but I bet that it was always called with the member,
not the aggregate, as the aggregate is either hard coded 
or was not needed...

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/library/shims/TestServiceShim.java
robin-aws: Just neglected to clean them up

.github/workflows/pull.yml
robin-aws: Definitely possible but not nearly as simple as you'd think :) I made the change as I already had to do something similar for Dafny (where we needed a lot more dynamic calculation)

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/Dafny.java
robin-aws: Yup, I just found there were cases of `Option<T>` and `Result<T, Error>` where `T` was driven by one of these aggregate shapes, so it was necessary to handle more cases in this method.

.github/workflows/pull.yml
```diff
@@ -5,36 +5,39 @@ on:
...
 
 jobs:
+  pr-populate-dafny-versions:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Populate Dafny versions list
+      id: populate-dafny-versions-list
+      run: echo "dafny-versions-list=['4.1.0', '4.3.0']" >> $GITHUB_OUTPUT
+    outputs:
+      dafny-version-list: ${{ steps.populate-dafny-versions-list.outputs.dafny-versions-list }}
```
texastony, line 15: Wow... GitHub needs to work on this...

robin-aws: And good luck if you want to pass values between WORKFLOWS :)

texastony (approved):

PR #304, opened 2023-12-01
feat: Start adding design docs

seebees:
Initial overloaded Readme

I will also add a few other diagrams.
We can break up this file as needed.


 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+
+DafnyImplementation --> SdkInterface
+SdkInterface --> DafnyImplementation
+
+```
+
+The goal is for Dafny developers
+to call native AWS SDKs.
+
+### Wrapped Library
```
lavaleri: I'm not sure how to improve this section, but it feels like a new reader won't get the point unless they already understand how all these parts work together and the motivations for needing to set things up this way.

How I understand this in my brain:
"I have a smithy-dafny library, and I want to test one of it's APIs" ->
"I could manually write tests per runtime, but that is expensive" ->
"I could write Dafny test against one of the Dafny interfaces, but then I'm not testing any of smithy-generated code" ->
"Thus, we wrote a WrappedLibrary feature, which allows you to call a smithy-dafny library within your dafny code" ->
"Write tests in Dafny, referencing the wrapped smithy-dafny library, to get integrations tests in every language"

We sort of hit on all of these in the text below, but I don't think it's written to guide new readers. Is the point of this design to document things for those with context, or to help new readers understand why things are set up the way they are?

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+Interface --> ToDafnyW
+
+DafnyTest --> WrappingInterface
+WrappingInterface --> DafnyTest
+
+```
+
+The goal is for Dafny developers
+to approximate native calls into Dafny libraries.
+`smithy-dafny` uses this feature to verify its features.
```
lavaleri: > verify its features.

I'm not sure what we mean here?

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+end
+
+UserCode --> Interface
+Interface --> UserCode
+```
+
+The goal is for native developers
+to call Dafny libraries with idiomatic types.
+
+### Wrapped AWS Services
```
lavaleri: Similar to below, it might be good to prime new readers with the problem this feature is solving. e.g. "If I am writing code in Dafny and want to make a call to an AWS service, how do I do that?"

 design/README.MD
lavaleri: I see that in a later section we _do_ have this narrative... Unsure from a doc perspective if the narrative should be brought up here, or whether any breadcrumb can be left here to say "wait this will make sense later I promise"

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+When developing `polymorph-X`
+the first thing to ask is:
+Does `smithy-X` exist?
+
+If it does exist then:
+`polymorph-X` SHOULD re-use as much `smithy-X` code-gen as possible.
+`polymorph-X` MUST use the same native types as `smithy-X`.
+
+If it does not exists then look at the AWS SDK in X:
+`polymorph-X` SHOULD use the same native types as the native AWS SKD in X runtime.
```
lavaleri:
```suggestion
`polymorph-X` SHOULD use the same native types as the native AWS SDK in X runtime.
```

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+Either way we need to test
+that these things are true
+and that the correct duals are connected together.
+
+It is much easier if we can write our tests in Dafny.
+Then we don't need to write new tests for each runtime.
+This is where [wrapping a library](#wrapped-library) comes in.
+
+We want the test in Dafny and we want to test the conversions.
+by building a project we can control the implementation.
```
lavaleri: I'm not sure what the purpose of this sentence is

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+By wrapping this library we can call the project from Dafny
+and pass through the type conversions.
+This sandwiches the conversions between two functions we control.
+Many of these are already written in the TestModels directory.
+
+For the simple types: string, number, et al
+the implementation can not just echo back values.
+This is because it is possible that the conversion function
+do not produce rational input in the middle.
+This is an edge case, but it is easy enough to verify.
```
lavaleri: I'm having trouble understanding this. Example would be useful

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+
+Each simple type has endpoints that:
+* returns back what is given.
+* takes will only succeed if pass a specific value.
```
lavaleri: ?

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+This sandwiches the conversions between two functions we control.
+Many of these are already written in the TestModels directory.
+
+For the simple types: string, number, et al
+the implementation can not just echo back values.
+This is because it is possible that the conversion function
+do not produce rational input in the middle.
+This is an edge case, but it is easy enough to verify.
+
+Each simple type has endpoints that:
```
lavaleri: Are these requirements for three different endpoints? Also I'm not sure if "endpoint" makes the most sense here.

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+We have one last subtlety to consider.
+Since the tests and implementation are both in Dafny
+we can clearly see the Dafny representation is reasonable.
+But this does not tell us that the native representation is reasonable.
+
+Consider a string.
+This string is represented under the hood as bytes
+and these bytes have an encoding.
+In `smithy-dafny` we represent this string as UTF8 bytes.
+If the conversion was to simply change the encoding to USC2
```
lavaleri: I'm having trouble understanding the example, as there are multiple conversions happening.
i.e. the ToDafny would go UTF8->USC2 and the ToNative would just USC2->USC2?

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+then all our tests would pass.
+But looking at the native string it would be drastically different.
+
+This is a degenerate example but should illustrate the point.
+At this time we do not have any such tests.
```
lavaleri: > such tests

What tests? Tests the ensure we don't do the above degenerative example?

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+This is because maintaining such coverage across multiple runtime
+is exceedingly difficult.
+It might be that a specification of such tests in conjunction with duvet
```
lavaleri: These tests would need to be hand-written in each runtime, yes?

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+`smithy-dafny` can take a smithy model
+and output the generated native code anywhere.
+A `runtimes` directory exists in a `smithy-dafny` project
+and inside that there is a directory for each native project.
+
+Looking at the TestModels directory you can see examples.
+The native project can be set up however is best for the native language.
+However the native project needs to bind the following things together
+
+1. `smithy-dafny` generate Polymorph code
```
lavaleri:
```suggestion
1. `smithy-dafny` generated Polymorph code
```

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+1. Translated Dafny implementation code
```
lavaleri: Elsewhere in the doc you use "transpiled"

 design/README.MD
lavaleri: Also, as a nit some of this polymorph code is Dafny, and thus needs to be transpiled itself somewhere. I think that's worth calling out.

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+that will expose native types and interfaces
+for your Smithy model.
+This creates a shim with native interfaces.
+These native interfaces are connected
+to the Dafny interface
+and takes care of type conversion back and forth.
+
+```mermaid
+---
+title: Library development
```
lavaleri: Nit: This diagram is not about the library development

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+---
+flowchart RL
+subgraph Library[Dafny Library]
+    direction RL
+    DafnyImplementation[Dafny implementation]
+end
+
+subgraph SDK
+    direction RL
+    SdkInterface[Dafny SDK Interface] --> ToNative[Convert input \nto native]
```
lavaleri: "Dafny SDK Interface" feels not quite right/not clear enough. It's the smithy-dafny generated Dafny interface of the SDK. Not sure how to make that clear, but also concise.

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+---
+title: Wrapped AWS service
+---
+flowchart RL
+subgraph Library[Dafny Library]
+    direction RL
+    DafnyImplementation[Dafny implementation]
+end
+
+subgraph SDK
```
lavaleri: Why not just call this the Wrapped SDK?

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+
+```mermaid
+---
+title: Wrapped Library
+---
+flowchart RL
+
+subgraph Wrapping[Wrapping Library]
+    direction RL
+    WrappingInterface[Dafny SDK Interface] --> ToNativeW[Convert input \nto native]
```
lavaleri: Similar comment for "Dafny SDK Interface" here.

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+that will convert the Dafny to native
+and then call the native implementation.
+
+```mermaid
+---
+title: Wrapped Library
+---
+flowchart RL
+
+subgraph Wrapping[Wrapping Library]
```
lavaleri: Would it be correct to call this the Wrapped Library instead?

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+    Interface[Public Interface] --> ToDafny[Convert input \nto Dafny]
+    ToDafny --> DafnyImplementation[Dafny implementation]
+    DafnyImplementation --> ToNative[Convert output \nto native]
+    ToNative --> Interface
+end
+
+ToNativeW --> Interface
+Interface --> ToDafnyW
+
+DafnyTest --> WrappingInterface
```
lavaleri: Should Dafny Test be it's own "Dafny implementation" subgraph?

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+    Native
+end
+
+Model --> Polymorph
+Polymorph --> Shim
+Native --> NativeTypes
+Polymorph --> ToNative
+Polymorph --> ToDafny
+Model --> Dafny
+Dafny --> DafnyTypes
```
lavaleri: Isn't it not only Native/DafnyTypes but the Native/Dafny APIs as well?

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+service --> Model
+
+Model --> types[ServiceTypes.dfy] & model[model.smithy]
+
+runtimes --> runtime
+
+runtime --> projectFile
+runtime --> RuntimeTests[tests]
+runtime --> Generated
+runtime --> Implementation
```
lavaleri: In my mind there are three things that need to be organized in each runtime in some way:
- Smithy->Dafny->runtime generated code (from model)
- Dafny->runtime generated code (from dafny/src)
- hand-written externs

It's not clear what "Generated" and "Implementation" mean here.

 design/README.MD
```diff
@@ -0,0 +1,421 @@
...
+```mermaid
+graph LR
+
+model[model.smithy] --> types[ServiceTypes.dfy] & Generated
+types[ServiceTypes.dfy] --> Index.dfy & ServiceOperations.dfy
+Index.dfy --> ServiceOperations.dfy
+Index.dfy --> Implementation
+Index.dfy --> DafnyTests[test] --> RuntimeTests[tests]
+Implementation --> RuntimeTests[tests]
+Generated --> Implementation
```
lavaleri: It seems like by "Generated" you mean "smithy->dafny generated code", and by "Implementation" you mean "Index.dfy dafny code that has been transpiled". Why doesn't the Operations.dfy lead anywhere?

I think we're drawing a line between what smithy-dafny is producing, and what the ESDK (e.g.) business logic is producing, but given how complex it all is, I think we need to clearly define what Dafny code where is important where, which I think means at least giving a name to the ESDK specific portion.

 design/detailed-diagrams.md
```diff
@@ -0,0 +1,686 @@
...
+
+  JsA -->|Dependency| JsB
+  JsC -->|Dependency| JsB
+end
+
+Dafny ---> Java
+Dafny ---> Net
+Dafny ---> Js
+```
+
```
lavaleri: This section is incomplete. Where does the previous "Now let us..." go?

 design/detailed-diagrams.md
```diff
@@ -0,0 +1,686 @@
...
+However organizing code required packages and dependencies.
+Looking at how the AWS Encryption SDK (ESDK)
+packages its dependencies
+we can see that having duplicate symbols
+for the `Crypto` library is problematic.
```
lavaleri: It took be a couple reads to see that this was saying "Look how we (correctly) have the ESDK dependencies expressed. Wouldn't it be really bad if we did it the bad way?" and not "Look how incorrect and bad this is".

 design/detailed-diagrams.md
```diff
@@ -0,0 +1,686 @@
...
+
+Dependencies are a problem for Dafny
+because by default Dafny includes all code
+into a single program.
+Dafny then translates all this code
+into a single program.
+This is illustrated in following diagram.
+The difficulty here is that when A and C
+try and exist in the same environment,
+then B symbols will be duplicated.
```
lavaleri: Should make clear here that B is another Dafny library

 design/detailed-diagrams.md
```diff
@@ -0,0 +1,686 @@
...
+
+subgraph Dafny
+  subgraph A["Package for A"]
+    ATypes["Dafny Types for A"]
+    AInterface["Dafny Interface for A"]
+    AImplementation["Dafny implementation of A"]
+
+    AImplementation --> AInterface
+    ATypes --> AInterface
+    ATypes --> AImplementation
```
lavaleri: What do these arrows mean? "Are/Is used by"?

 design/detailed-diagrams.md
```diff
@@ -0,0 +1,686 @@
...
+
+  ATests ==>|Dafny Compilation| NetATests
+  NetATests -->|Testing| NetAFromDafny
+end
+```
+
+### Dependencies
+
+This is a description of how Smithy Dafny
+handles dependencies.
```
lavaleri: It would help to explain exactly what we are seeing here. It looks like we have two packages, A and B, both are smithy-dafny projects, and B depends on A.

 design/detailed-diagrams.md
```diff
@@ -0,0 +1,686 @@
...
+---
+title: How Smithy Dafny handles dependencies
+---
+graph LR
+
+subgraph Smithy
+  AModel["Model for A"]
+  BModel["Model for B"]
+
+  BModel --> AModel
```
lavaleri: depends on?

 design/detailed-diagrams.md
```diff
@@ -0,0 +1,686 @@
...
+    NetAInterface
+    NetATypeConversion["To/From Dafny"]
+
+    NetATypeConversion --> NetAInterface
+    NetAFromDafny --> NetAInterface
+    NetATypes --> NetAInterface
+    AModel -.->|CodeGen| NetATypes
+    AModel -.->|CodeGen| NetAInterface
+    AModel -.->|CodeGen| NetATypeConversion
+    A ==>|Dafny Compilation| NetAFromDafny
```
lavaleri: This subgraph isn't making sense to me

 design/detailed-diagrams.md
```diff
@@ -0,0 +1,686 @@
...
+
+    ATypes ~~~ NetAInterface
+  end
+
+  NetB["B<sub>.net</sub>"]
+  B ==>|Dafny Compilation| NetB
+
+  NetB --> NetAFromDafny
```
lavaleri: it's not clear, but this is saying that the "internals" of the .NET B project call into "internals" of the .NET A project?

 design/detailed-diagrams.md
```diff
@@ -0,0 +1,686 @@
...
+  BModel -.->|CodeGen| NetB
+
+end
+```
+
+### Wrapped AWS Services
```
lavaleri: Also having a hard time groking this one. Using a concrete example would help I think.

 design/detailed-diagrams.md
```diff
@@ -0,0 +1,686 @@
...
+
+  NetATests
+
+  ATests ==>|Dafny Compilation| NetATests
+  NetATests --> NetAFromDafny
+end
+
+```
+
+### Extern Library Development
```
lavaleri: An example here as well.

 design/detailed-diagrams.md
```diff
@@ -0,0 +1,686 @@
...
+  NetATests --> NetAExtern4
+
+end
+
+%% Formatting
+Smithy ~~~~~ Dafny
+
+```
+
+### Testing/Verification
```
lavaleri: Example pls

 design/README.MD
seebees:
```suggestion
`smithy-dafny` uses the wrapped library feature
to test the correctness of polymorph-runtime features.
```

 design/README.MD
```diff
@@ -0,0 +1,422 @@
...
+to the Dafny types produced by `smithy-dafny`.
+
+```mermaid
+---
+title: Generated Code
+---
+flowchart LR
+
+subgraph Project
+    Shim
```
seebees:
```suggestion
    NativeInterface
```

 design/README.MD
```diff
@@ -0,0 +1,422 @@
...
+    NativeTypes
+    ToNative
+    ToDafny
+    DafnyTypes
```
seebees:
```suggestion
    DafnyTypes
    DafnyInterface
```

 design/README.MD
seebees: We want to test the type conversions.
We also want to write these tests in Dafny
so that we can use these tests for every runtime.

To test the type conversions,
you would need to write them in every language.
But we do not want to write the tests in every language.
If we write a test and implementation in Dafny
we can run the conversions between these two
Dafny methods and test the type conversion.

 design/README.MD
seebees: Yes.

 design/README.MD
seebees: Got it

 design/README.MD
seebees: took a stab

 design/README.MD
seebees: an example follows. the string encoding.

 design/README.MD
seebees: fixed

 design/README.MD
seebees: changed to smithy operations.

 design/README.MD
seebees: yes. Updated.

 design/README.MD
seebees: changed

 design/README.MD
seebees: took a try

 design/README.MD
seebees: No, this library is wrapping the "Native Library" not the other way round.

 design/README.MD
seebees: Not really I think. Like all the tests together are organized, but each one is just a test.
The goal of doing all this is to make the tests simple.

 design/README.MD
seebees: Hum, yes. The interfaces are inside the Types file.
Added an additional box.
As that distinction just confuses things.

 design/README.MD
seebees: added a note

 design/README.MD
seebees: The Operations.dfy are included from Index.dfy.

 design/detailed-diagrams.md
seebees: better

 design/detailed-diagrams.md
seebees: done

 design/detailed-diagrams.md
seebees: Yes, added a legend

 design/detailed-diagrams.md
seebees: done

 design/detailed-diagrams.md
seebees: Yes added legend

 design/detailed-diagrams.md
seebees: Changed to KMS.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+that can safely evolve
+maintaining both forward and backwards compatibility.
+
+Dafny is a programming language with formal verification features.
+Dafny code is translated to higher-level languages
+to provide interoperability for your verified code
+like Java or C#.
+These native representations are then executable.
+This means that a single Dafny program
+can be translated into multiple interoperable programs.
```
robin-aws:
```suggestion
can be translated into multiple equivalent programs.
```

"Interoperable" makes sense for something like the ESDK, because you can talk about using the output of one instance as the input for another, but I don't think interoperability between the equivalent programs will apply to all polymorphed libraries.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+to the obvious string type in each translated language.
+
+However a native `string` may be UTF8, UTF16/USC2, or UTF32 
+depending on the language.
+Numbers are similarly complicated.
+Some languages have different sized numbers,
+while others have either a different default or a single size.
+There are also complicated types like streams.
+All these languages have equivalent
+but not congruent types to each other.
```
robin-aws: "congruent" doesn't immediately resonate for me in this context. Are you trying to point to the fact that for e.g. `string` in one language may allow values that it doesn't in others?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
+## Overview
```
robin-aws: This section is biasing a bit too much for just the polymorphing use case IMO. It's probably worth framing this as "Like any Smithy-based code generator, smithy-dafny generates Dafny code for Smithy interfaces, but there's more to it than that because Dafny supports multi-targeting" etc.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Smithy-Dafny modes
+
+### Library
+
+`smithy-dafny` supports `@localService` development.
+This is where the implementation is locally executing code.
+Rather than a remote service like most AWS Service Smithy models.
```
robin-aws:
```suggestion
This is where the implementation is locally executing code,
rather than a remote service like most AWS Service Smithy models.
```

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+flowchart RL
+subgraph Library[Native Library]
+    direction RL
+    Interface[Public Interface] --> ToDafny[Convert input \nto Dafny]
+    ToDafny --> DafnyImplementation[Dafny implementation]
+    DafnyImplementation --> ToNative[Convert output \nto native]
+    ToNative --> Interface
+end
+
+UserCode --> Interface
```
robin-aws:
```suggestion
UserCode[User Code] --> Interface
```

Assuming that works, first time using mermaid :)

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+`smithy-dafny` also generates polymorph code
+that will expose native types and interfaces
+for your Smithy model.
+This creates a shim with native interfaces.
+These native interfaces are connected
+to the Dafny interface
+and takes care of type conversion back and forth.
+
+```mermaid
```
robin-aws: Love the diagrams in general!

Is it possible to align the two Convert boxes vertically? That would help the intuition a lot.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+Interface --> UserCode
+```
+
+The goal is for native developers
+to call Dafny libraries with idiomatic types.
+
+### Wrapped AWS Services
+
+`smithy-dafny` also supports wrapping existing AWS Services
+with existing implementations that have a native implementation.
```
robin-aws:
```suggestion
`smithy-dafny` also supports creating Dafny AWS Service clients
by wrapping existing native client implementations.
```

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+    `N == ToNative(ToDafny(N))`
+
+### Conventions used in this document
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Smithy-Dafny modes
+
+### Library
```
robin-aws: I'd flip the order of these two sections since "Wrapped AWS Services" is the more typical smithy-runtime use case. Then you can say "`smithy-dafny` also support local services..."

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+Again there is an `abstract module` that you refine
+with a `extern` that returns a native SDK instance
+wrapped with the Dafny polymorph shim.
+Ideally these `extern` would be generated
+but the constructors for the SDKs
+are not described in Smithy.
+
+In this diagram we zoom in on
+only the Dafny implementation.
```
robin-aws: I think this might confuse rather than clarify. You're probably imagining "Dafny Library" as a more complicated polymorphing library itself, but if you just set it up as any Dafny codebase that needs to call an AWS service, the reader will already be focused on just the Dafny code that does that I think.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+- Native type or native interface:
+    A type/interface in a runtime,
+    e.g. Java `String`
+- Smithy-runtime:
+    The project used to generate native types
+    for Smithy shapes.
+- Polymorph or polymorph-runtime:
+    The transformation layer between Dafny types
+    and native types,
+    e.g. polymorph-java.
```
robin-aws: +100 on starting to make this distinction in general.

There's still a point of confusion since the `smithy-dafny` repository contains both "Dafny code generation for Smithy shapes" (what you're calling `smithy-dafny` here) and "translating code generation between Dafny and native types for Smithy shapes" (what you're calling `polymorph-<runtime>` here)

Maybe we can lean towards the terminology idioms other Smithy projects are using at a lower level. There's Smithy build plugins like `typescript-client-codegen` vs. `typescript-server-codegen`, and in our case we probably want `dafny-client-codegen` vs. `dafny-library-codegen`. What we want is labels for the common pieces that those plugins are built on top of.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+## Polymorph development
+
+When developing `polymorph-X`
+the first thing to ask is:
+Does `smithy-X` exist?
+
+If it does exist then:
+`polymorph-X` SHOULD re-use as much `smithy-X` code-gen as possible.
+`polymorph-X` MUST use the same native types as `smithy-X`.
```
robin-aws: Good to include at least one link to another `smithy-X` repo here as an example.

robin-aws:
```suggestion
If it does exist, then
`polymorph-X` SHOULD re-use as much `smithy-X` code-gen as possible, and
`polymorph-X` MUST use the same native types as `smithy-X`.
```

Or expand into multiple sentences.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+title: Type conversion dual
+---
+flowchart LR
+NativeTypes --> ToDafny --> DafnyTypes
+DafnyTypes --> ToNative --> NativeTypes
+```
+
+However there may be many to one relationships.
+Perhaps `string` and `char[]`.
+These two would convert to `ValidUTF8Bytes` or `seq<uint8>`.
```
robin-aws: I'd pick an example language and be very precise. "For example in Java, either a `java.lang.String` or a `char[]` is a common way of providing string data. Either one should be converted to the same idiomatic representation of string in Dafny (either `ValidUTF8Bytes` or `string` or...)."

That's not actually the best example since `java.lang.String` is by far the most common. Rust might be better since it has three different flavors of strings. Also strings are an annoying case for Dafny right here since there are multiple choices, so you're actually jumping to "many to many" right away.

Perhaps blobs are even better in general: interfaces very commonly accept either a ByteBuffer or a byte[] + start + length triple in Java.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+Also reference types may not be exactly equal.
+While a native array would convert to a `seq`
+when converting back the new array
+will have the same elements in the same order
+but may not be exactly the same reference object.
+We can refer to this as `SemanticallyEqual`.
```
robin-aws: Or perhaps "EqualValues"?

One thing to mention loudly at some point is the fact that Smithy shapes all have only value semantics. Even a Smithy stream is really just a fixed sequence you access incrementally, not a pointer to a mutable thing in the external environment. Services and Resources are values that act like references.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+The spirit of this can be expressed
+```dafny
+lemma ConversionsCorrectlyInvert<D, N>(ToDafny<D,N>, ToNative<N, D>, SemanticallyEqual)
+  ensures forall d, n ::
+    SemanticallyEqual(ToNative(d), n) <==> SemanticallyEqual(ToDafny(n), d)
+```
```
robin-aws: FWIW if your `N` is a Dafny extern declaration for the native type, this could at least actually type check even if it's obviously not provable. When we have Dafny standard libraries for interop and `ToDafny` and `ToNative` are actually Dafny functions, you could imagine somehow generating tests from this specification. :)

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+Either way we need to test
+that these things are true
+and that the correct duals are connected together.
+
+It is much easier if we can write our tests in Dafny.
+Then we don't need to write new tests for each runtime.
+This is where [wrapping a library](#wrapped-library) comes in.
+
+We want the test in Dafny and we want to test the conversions.
+by building a project we can control the Dafny implementation.
```
robin-aws:
```suggestion
By building a project we can control the Dafny implementation.
```
And I think you mean "by building a sample, test project"?

 design/README.MD
robin-aws: I'd still reorder things so you start with the problem and an example, and then say "to fix this, each simple type has smithy operations that..."

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+and these bytes have an encoding.
+In `smithy-dafny` we represent this string as UTF8 bytes.
+If the conversion was to simply change the encoding to USC2
+then all our tests would pass.
+But looking at the native string it would be drastically different.
+
+This is a degenerate example but should illustrate the point.
+At this time we do not have any such degenerate tests written.
+This is because maintaining such coverage across multiple runtime
+is exceedingly difficult.
```
robin-aws: More time-consuming than technically difficult though?

 design/README.MD
robin-aws: Same here. Is the problem that the Dafny tests don't detect that the native representation is not idiomatic for that language, and the proposed solution to write manual tests in the language that directly assert the representation is idiomatic?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+It might be that a specification of such tests in conjunction with duvet
```
robin-aws: I'd either link to Duvet and give a bit more background on why it's relevant, or not mention it.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+might be a possible solution.
+
+## Project Management
+
+Neither Dafny nor `smithy-dafny` provide project management features.
+They are not terribly opinionated
+about how native project should be set up.
```
robin-aws: Really? :) Seems to me like the shared makefile is very opinionated about how everything is laid out.

I think this is just another issue with the ambiguity of `smithy-dafny` though - the repo on the whole is opinionated because it includes the shared makefile, but you're trying to say the code generation tools don't care (which I think needs to change, since we should be generating more than just the code from the Smithy models)

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+`smithy-dafny` can take a smithy model
+and output the generated native code anywhere.
```
robin-aws: I think we need to pull in the difference between the CLI and the Smithy build plugin at this point.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+A `runtimes` directory exists in a `smithy-dafny` project
+and inside that there is a directory for each native project.
+
+Looking at the TestModels directory you can see examples.
+The native project can be set up however is best for the native language.
+However the native project needs to bind the following things together
+
+1. `smithy-dafny` generated Polymorph code
+1. Translated Dafny implementation code
```
robin-aws: Only for libraries and not for SDK projects though, right?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+1. Translated Dafny test code
+1. Native `smithy-dafny` dependencies
+    * Dafny runtime
+    * Dafny conversion libraries
+    * Dafny project dependencies
+    * Native dependencies -
```
robin-aws:
```suggestion
    * Native dependencies
```

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+    * Dafny conversion libraries
+    * Dafny project dependencies
+    * Native dependencies - 
+
+In the TestModels directory there is a SharedMakefile
+that can be used to provide useful targets for Dafny projects.
+
+Here is an example project structure.
+
+```mermaid
```
robin-aws: This one shouldn't be a diagram, just ascii art. e.g. the directory structure snippet from https://github.com/smithy-lang/smithy-typescript?tab=readme-ov-file#using-smithy-typescript-with-gradle

robin-aws (changes requested):
Made it through the main README, will come back to the detailed diagrams. Awesome stuff so far.

 design/README.MD
robin-aws, line 1: top-level nit: just noticed the Smithy guide specifies `designs` rather than `design`: https://smithy.io/2.0/guides/building-codegen/creating-codegen-repo.html#codegen-repo-layout

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+from any given executable implementation.
+By focusing on the abstract interface
+Smithy helps you build an interface
+that can safely evolve
+maintaining both forward and backwards compatibility.
+
+Dafny is a programming language with formal verification features.
+Dafny code is translated to higher-level languages
+to provide interoperability for your verified code
+like Java or C#.
```
andrewbanchich: > Dafny code is translated to higher-level languages
to provide interoperability for your verified code
like Java or C#.

seems a bit confusingly worded. maybe:

> Dafny code is translated to higher-level languages like Java or C#
to provide interoperability with your verified code

?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+Numbers are similarly complicated.
+Some languages have different sized numbers,
+while others have either a different default or a single size.
+There are also complicated types like streams.
+All these languages have equivalent
+but not congruent types to each other.
+
+In this project Dafny and Smithy are combined
+to generate concrete interfaces in Dafny
+and supported Dafny translation targets.
```
andrewbanchich: translation == transpilation?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+These generated interfaces all use the same translated Dafny implementation
+and expose native interfaces and types for each language.
```
andrewbanchich: nit: IMO it might be clearer to word it like

> this allows a single implementation in Dafny to generate native interfaces in other languages

since it begins with what seems like the most important part and is structured to match the process of using the tool (write Dafny -> generate native libraries -> use libraries).

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+generate interfaces that can be used like this:
+1. Dafny -> `var hello: seq<char> :- greet();`
+1. Java -> `String hello = greet();`
+1. C# -> `string hello = greet();`
+1. Python -> `hello = greet()`
+1. Rust -> `let hello: String = greet();`
+
+## Definitions
+
+- Runtime or native runtime:
```
andrewbanchich: i'm new to the project so might be lacking some context - but isn't "runtime" a super overloaded term? why not just call them "transpilation targets"?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+    e.g Java or C#.
+    This also may include additional context
+    like language version or other versions/features.
+- Native type or native interface:
+    A type/interface in a runtime,
+    e.g. Java `String`
+- Smithy-runtime:
+    The project used to generate native types
+    for Smithy shapes.
+- Polymorph or polymorph-runtime:
```
andrewbanchich: are these synonymous? if so, any thoughts on just standardizing on "polymorph" so that usage of "polymorph-runtime" doesn't imply it's something different?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+    The transformation layer between Dafny types
+    and native types,
+    e.g. polymorph-java.
+- Shim:
```
andrewbanchich: even though it's a shim, would a less general term help clarify its purpose?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+## Smithy-Dafny modes
+
+### Library
+
+`smithy-dafny` supports `@localService` development.
+This is where the implementation is locally executing code.
+Rather than a remote service like most AWS Service Smithy models.
+
+In this mode `smithy-dafny` will produce Dafny types and interfaces
+for a Smithy model in Dafny.
```
andrewbanchich: isn't that what smithy does in general, even without `@localService`? also, is this sentence redundant (produce Dafny types ... in Dafny)?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+This creates an `abstract module` in Dafny you can refine
+with a concrete implementation.
```
andrewbanchich: i'd love some more details on this. so far i've understood the previous explanations, but i don't know what this means.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+`smithy-dafny` also generates polymorph code
+that will expose native types and interfaces
+for your Smithy model.
```
andrewbanchich: this part is a bit unclear to me. e.g. "generates polymorph code" - "polymorph" already refers to a tool which generates code, so does "generate polymorph code" generate code which generates code? also, i feel like "native types" seems ambiguous in the context of this project, since i might interpret it to mean, e.g.

1. `Rc<::simple_boolean_dafny::r#_simple_dtypes_dsmithyboolean_dinternaldafny_dtypes::GetBooleanInput>` - native in the sense that it's Rust
2. `Option<bool>` - native in the sense that it's Rust and has nothing to do with Dafny

just a suggestion, but maybe we could say "idiomatic" type / interface for example 2?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+This creates a shim with native interfaces.
+These native interfaces are connected
+to the Dafny interface
```
andrewbanchich: just fyi - this part took me some time to digest, but that was probably due to my unfamiliarity with Dafny. e.g. "we're calling Dafny code from Rust? how?" for me, as a newcomer, it would have been clearer to have a term that clarified it was the "native but non-idiomatic code based on Dafny's typesystem / assertions" (if that's the right way to think of it).

 design/README.MD
andrewbanchich: also, reading this leads me to more questions, like - does overriding an impl with an existing native language impl mean it's no longer formally verified? would the assertions Dafny makes for non-existing clients change behavior between those and the native one?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+`smithy-dafny` also supports wrapping existing AWS Services
+with existing implementations that have a native implementation.
+
+The problem we are trying to solve here is:
+I am writing code in Dafny
+and I want to call an AWS service,
+how do I do that?
+
+In this mode `smithy-dafny` will produce Dafny types and interfaces
+so that Dafny code can call the native implementation.
```
andrewbanchich: isn't client generation something which Smithy does generally? i'm unclear on how this would work. "wrapping a client" sounds like runtime behavior, but Dafny doesn't have a runtime, right? e.g. if Java is a transpilation target for Dafny and `smithy-dafny` is configured to wrap the native Java client in order to make calls to some AWS service, then what happens for non-Java transpilation targets?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+`smithy-dafny` also generates polymorph code
+so that these Dafny types and interfaces
+are connected to the native interface.
+Along with all the required type conversions.
+
+Again there is an `abstract module` that you refine
+with a `extern` that returns a native SDK instance
```
andrewbanchich: thoughts on adding "abstract module" and "extern" to the `Definitions` section?

 design/README.MD
andrewbanchich: is the term "library" too generic here? if this is specifically about `@localService` then maybe call the section `Local Service` since, from what i understand, this is about 100% client side code with no HTTP requests. or, at least, an HTTP request is not the sole abstraction like with AWS service SDKs.

 design/README.MD
andrewbanchich: `Generate Dafny AWS SDK clients` maybe? `smithy-rs` worded it like this: https://smithy-lang.github.io/smithy-rs/design/overview.html#code-generation

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+The goal is for Dafny developers
+to call native AWS SDKs.
+
+### Wrapped Library
+
+`smithy-dafny` can wrap an existing `smithy-dafny` library.
+This is useful for integration testing.
+
+When you consider the first two use cases
```
andrewbanchich: use cases == modes? maybe this could say e.g. "when you consider `smithy-dafny` use cases like local services and wrapped AWS services, it's clear that..." and / or maybe the `Smithy-Dafny modes` section could be renamed to `Smithy-Dafny use cases`?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+it is clear that you need tests
+to verify that these features are correct.
```
andrewbanchich: > When you consider the first two use cases

is there something special about local clients and wrapped AWS services here, or is this just generally "you should write unit tests for your libraries" regardless of specifics?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+The first thought is be to write tests in each runtime.
```
andrewbanchich: typo

> Your first thought might be to write tests in each runtime, but this is very expensive.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+But this is very expensive.
```
andrewbanchich: expensive in what way? to maintain? to run?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+Also keeping track of all these tests across different runtimes is difficult.
+You don't want to have tests only in some runtimes but not others.
```
andrewbanchich: you also wouldn't be able to implement all tests in all languages. e.g. you may want to test that a value is never null in Java, but you can't write that test against idiomatic Rust code since null doesn't exist in the language; if you want a never-null `String` then just write `String`. so tests would all have to be customized around each language's design to fill in the blanks for what their type systems do not enforce.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+If you write the tests in Dafny then you only need to write the test once.
+But if you natively write these tests in Dafny
+when you execute them
+they will only test the Dafny code.
```
andrewbanchich: another newcomer-to-Dafny question: if Dafny doesn't have a runtime, only transpilation targets, then how do native Dafny tests run?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+Because they will not exercise the type conversions.
+
+So we want to write the tests in Dafny so that they are interoperable
+but we need to exercise the native code.
+By wrapping the whole project
+we force an additional type  conversion.
```
andrewbanchich: extra space

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+The first thought is be to write tests in each runtime.
+But this is very expensive.
+Also keeping track of all these tests across different runtimes is difficult.
+You don't want to have tests only in some runtimes but not others.
+
+If you write the tests in Dafny then you only need to write the test once.
+But if you natively write these tests in Dafny
+when you execute them
+they will only test the Dafny code.
+Because they will not exercise the type conversions.
```
andrewbanchich: > Because the generated idiomatic APIs all have custom, language-specific logic for converting from their unidiomatic, Dafny-specific counterparts, tests written in native Dafny would not be sufficient at making assertions about the idiomatic APIs.

is this corrrect?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+So we want to write the tests in Dafny so that they are interoperable
```
andrewbanchich: instead of "interoperable" would it be clearer to say "standardized"? "uniform"?

 design/README.MD
andrewbanchich: not sure i understand this

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+but we need to exercise the native code.
+By wrapping the whole project
+we force an additional type  conversion.
+
+Writing tests in Dafny means that
+your testing is now also interoperable.
+A Dafny tests can check multiple runtimes.
+But a simple test would only check the Dafny interfaces.
```
andrewbanchich: seems like this just repeats previous section, but doesn't clarify how the issue was solved or how a Dafny test can check multiple runtimes

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+For more details on how this works
+see the [polymorph development](#polymorph-development) section blow.
+
+This feature is used to test `smithy-dafny` on itself
+in the TestModels.
```
andrewbanchich: dogfooding? if so, might help to use that term.

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+This features is used by the Encryption SDK
```
andrewbanchich: > This feature is used to test `smithy-dafny` on itself
in the TestModels, and is used by the Encryption SDK...

andrewbanchich: what is the Encryption SDK? it hasn't been mentioned until now. i see its mentioned later in an example. it sounds like this is just giving an example of a client which is doing this, but not sure if adding this extra section helps to clarify anything.

 design/README.MD
andrewbanchich: >`polymorph-X` MUST use the same native types as `smithy-X`

is this something which could be enforced programmatically?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+
+When developing `polymorph-X`
+the first thing to ask is:
+Does `smithy-X` exist?
+
+If it does exist then:
+`polymorph-X` SHOULD re-use as much `smithy-X` code-gen as possible.
+`polymorph-X` MUST use the same native types as `smithy-X`.
+
+If it does not exists then look at the AWS SDK in X:
```
andrewbanchich: typo

> If it does not exist

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+Consider a string.
+This string is represented under the hood as bytes
+and these bytes have an encoding.
+In `smithy-dafny` we represent this string as UTF8 bytes.
+If the conversion was to simply change the encoding to USC2
+then all our tests would pass.
+But looking at the native string it would be drastically different.
+
+This is a degenerate example but should illustrate the point.
+At this time we do not have any such degenerate tests written.
```
andrewbanchich: degenerate == specific, or just an example of a larger issue? could we use more common terminology?

 design/README.MD
```diff
@@ -0,0 +1,456 @@
...
+This is because maintaining such coverage across multiple runtime
+is exceedingly difficult.
+It might be that a specification of such tests in conjunction with duvet
+might be a possible solution.
+
+## Project Management
```
andrewbanchich: would `Project structure` be clearer? i initially interpreted this to mean sprint planning / Gantt charts.

 design/README.MD
seebees: I'm trying to distinguish between similar and congruent in terms of geometry. Like the difference between similar and congruent triangles.

When we are sloppy we talk about things bing the "same" but what we really mean is that they are similar. So every language has an Array, but not all Arrays are exactly the same.

 design/README.MD
seebees: Yes, Dafny calls it translation.

 design/README.MD
seebees: Each transpilation targets would need to have its own way to provide a native SDK to be wrapped.

 design/README.MD
seebees: No just that no matter how much you prove, you always need some test.

 design/README.MD
seebees: The are translated into a native runtime and executed.

 design/README.MD
seebees: Not really that I could see. Like you could write some tests that make sure that the native types `polymorph-X` types produce are the same as `smithy-X` but this is something you would do as writing `polymorph-X`.

 design/README.MD
seebees: `overly simplified`

 design/README.MD
seebees: I tried, but I can't make it work :(

 design/README.MD
seebees: No everyone always has to do this :(

robin-aws (approved):
I'm super happy to get this added, and although I do want to take another pass at it and/or add more detail here or somewhere else, it's just so much better to have merged than sitting here. In that spirit I'm gladly approving even though I will likely cut a PR to improve even further relatively soon.

Thanks so much for taking the time on this!

Issue #133, opened 2023-02-06
Generate constructors with default assignments to non-required fields


codyroux:
The title sums it up: assigning `None` to non-required fields by default would avoid some pain when actually creating members of these records. I think this is the relevant Dafny section: https://dafny.org/dafny/DafnyRef/DafnyRef#2146-formal-parameters-and-default-value-expressions

ajewellamz:
Fixed. https://github.com/smithy-lang/smithy-dafny/pull/305
Leaving this open, because it needs coverage in a Test Model.

PR #305, opened 2023-12-04
feat: add default value None for optional members

ajewellamz:
*Issue #, if available:* 133

*Description of changes:*

I tested this on the MPL, and there seem to be other changes at the head of smithy-dafny that break the MPL, but these changes seem to be correct.


texastony (approved):
Approving, but it concerns me that there was no change to any Test Model.
I suggest leaving #133 open, with a "Needs Tests" label.

(If we do not have such a label or an equivlent, we should make one.)

PR #306, opened 2023-12-11
feat(TestModels): ExternV2

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

New Extern system TestModel.

This PR is primarily for `replaceable` modules. The `IndexNet.dfy` file shows how you can generate code for a particular language.

I also added new Makefile targets for `v2` extern modules. There are pretty useless at the moment. The intention was to use these to run `dafny translate`, but there are some issues with that:
* `dafny translate` cannot be used to generate tests. Tracking in https://github.com/dafny-lang/dafny/issues/4888
* `dafny translate` would require some work on the Smithy-Dafny side to not generate `_Compile` suffixes, or to only generate it in some cases. I'm not sure what that would look like right now; this depends on whether we need to use `dafny translate` to access `outer-module`.

We will pick these issues up again as part of https://sim.amazon.com/issues/CrypTool-5259 once `outer-module` is ready, or a replacement for it is ready. I'll leave these in so we don't have to recreate the Makefile structure for them.

Questions for reviewers:
- [ ] Does the module replacement system of (generic Index) -> (namespace-syntax-specific Index) -> (language-specific Index) make sense? 
- [ ] Do we like having one/two Concrete `Index` modules that includes ALL `replaces` for all `replaceable` modules? Or do we want to slice these `replaces` up more?

Other notes:

Two hefty TODOs remain for this TestModel (Java implementation, `outer-module` usage). These are blocked as of today. These are documented in https://sim.amazon.com/issues/CrypTool-5259


TestModels/ExternV2/README.md
```diff
@@ -0,0 +1,41 @@
+# SimpleExternV2
+
+This project tests implementing a [dafny extern](https://homepage.cs.uiowa.edu/~tinelli/classes/181/Papers/dafny-reference.pdf#15) using the "V2 Extern" system.
+This models and its shapes are identical to operations in the `Extern` TestModel, only with "V2" suffixes appended.
+
+This V2 extern system uses the `replaceable` and `replaces` keywords in conjunction with the `outer-module` compile flag to allow a module to have idiomatic extern names in each target language.
+
+Under this V2 system, manually-written Dafny source code should mark `module`s that are intended to be `extern` as `replaceable`.
```
ShubhamChaturvedi7: I think we should clarify here that by `extern` we mean having the module under a specific path, and not have an extern implementation.

TestModels/ExternV2/src/IndexDotNamespaced.dfy
```diff
@@ -0,0 +1,5 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+include "Index.dfy"
+
+module {:extern "simple.dafnyexternv2.internaldafny"} UnderscoreNamespacedSimpleExternV2 replaces SimpleExternV2 {}
```
ShubhamChaturvedi7: Typo: I think you meant `DotNamespacedSimpleExternV2` and not `Underscore..`.

ShubhamChaturvedi7 (commented):
Should we track the TODO work somewhere?

lucasmcdonald3:
> Should we track the TODO work somewhere?

2 TODOs:
* Python: No need to track, I'll implement this when I merge in Python codegen.
* outer-module. I'll revisit this as part of this PR. Let me take a second pass at this when I implement Java ExternV2.

TestModels/ExternV2/README.md
lucasmcdonald3: How about this:

```
Under this V2 system, a Dafny developer should not mark a module as `extern` if the path to the extern module depends on language-specific features. (e.g. an `:extern "sample.namespace.MyExternModule"` makes sense for Java externs, but the `.`s will break Python externs.)
Instead of marking this module as `extern`, the developer should mark the module as `replaceable`.
Then, they should create a new module that `replaces` the `replaceable` module. The developer should add the `extern` attribute to this new module. This module's `extern` attribute will define a language-idiomatic path syntax.
For each path syntax, the developer should create a separate module that `replaces` the `replaceable` module.
```

texastony:
> outer-module is not available on the legacy dafny CLI, but some of the CLI arguments Polymorph (and MPL) use with legacy CLI aren't available on dafny translate. We need to reconcile these somehow.

Do we have an issue/ticket with Danfy/ARG about this yet?

.github/workflows/test_models_net_tests.yml
```diff
@@ -48,6 +48,14 @@ jobs:
         ]
         dotnet-version: [ '6.0.x' ]
         os: [ ubuntu-latest ]
+        dafny-version: ${{ inputs.dafny }}
+        exclude:
+          # ExternV2 is only supported on Dafny >4.4.0 (unreleased; only on nightly as of commit)
```
texastony: @seebees I thought you said Dafny had already released this?
If Dafny has not released this, how can we use it for the ACCP work?

lucasmcdonald3:
> Do we have an issue/ticket with Danfy/ARG about this yet?

No, I was able to reconcile these. It looks like the legacy options that were missing are obsolete with `dafny translate`. These were `spillTargetCode`, `compile`, and `useRuntimeLib`. I'll push a commit with the corresponding `translate` build target.

lucasmcdonald3:
> > Do we have an issue/ticket with Danfy/ARG about this yet?
> 
> No, I was able to reconcile these. It looks like the legacy options that were missing are obsolete with `dafny translate`. These were `spillTargetCode`, `compile`, and `useRuntimeLib`. I'll push a commit with the corresponding `translate` build target.

Just kidding! This is actually an issue for test generation.
Issue: https://github.com/dafny-lang/dafny/issues/4888

.github/workflows/test_models_dafny_verification.yml
```diff
@@ -26,6 +26,7 @@ jobs:
           TestModels/Errors,
           TestModels/Extendable,
           TestModels/Extern,
+          # TestModels/ExternV2,
```
texastony:
```suggestion
          # TestModels/ExternV2, # Verified with Dafny 4.4.0 only
```

.github/workflows/test_models_net_tests.yml
```diff
@@ -14,6 +14,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        dafny-version:
+          - ${{ inputs.dafny }}
```
texastony, line 18: Nit: It would be nice to be consistent with our GitHub Workflow Variables.

TestModels/ExternV2/README.md
```diff
@@ -0,0 +1,67 @@
+# SimpleExternV2
+
+This project tests implementing a [dafny extern](https://homepage.cs.uiowa.edu/~tinelli/classes/181/Papers/dafny-reference.pdf#15) using the "V2 Extern" system.
+This models and its shapes are based on to operations in the `Extern` TestModel.
```
texastony:
```suggestion
The model and its shapes are based on to operations in the `Extern` TestModel.
```

TestModels/ExternV2/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+The differences between these two TestModels are:
+* V2 suffixes are appended to classes and operations
+* The NET implementation of the `simple.dafnyexternv2.internaldafny` extern contains an extern method that is not present in other languages (as a demonstration; see below)
+* The `ExternV2Constructor` is moved to `simple.dafnyexternv2.internaldafny.ExternV2Constructor` 
+
+This TestModel also demonstrates a "V2" extern system.
```
texastony: Suggestion: Remove repeat. This is stated on line 3.
```suggestion
```

TestModels/ExternV2/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+The "V1" extern system applies an `{:extern "..."}` attribute to a module to signal an extern.
+This worked for .NET and Java, where this extern string was valid in both languages.
+However, this does not work for Python and Go, where this string is not applicable.
+
+The "V2" extern system is intended to allow per-language extern strings.
+This system uses the `replaceable` and `replaces` keywords
+in conjunction with the `outer-module` compile flag
+to allow a module to have idiomatic extern names in each target language.
```
texastony, line 19: Issue/Blocking/Question: I thought `outer-module` did not work,
and we were removing it for just `replaceable`?

TestModels/ExternV2/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Under this V2 system, a Dafny developer should not directly mark a module as `extern` if the path to the extern module depends on language-specific features.
```
texastony, line 21: Question/Blocking: 
Is it language-specific features? Like only Java can use ACCP.
Or is it package naming? Like only Java should use `software.amazon.cryptography`.

Or, more confusingly, is it both?

TestModels/ExternV2/README.md
```diff
@@ -0,0 +1,67 @@
...
+(e.g. an `{:extern "sample.namespace.MyExternModule"}` makes sense for Java externs, but the `.`s will break Python externs.)
+
+Instead of marking this module as `extern`, the developer should mark the module as `replaceable`.
+Then, they should create a new module that `replaces` the `replaceable` module.
+The developer should add the `extern` attribute to this new module.
+This module's `extern` attribute will define a namespace-path-syntax-specific extern name (e.g. `.`s vs `_`s).
+For each path syntax, the developer should create a separate module that `replaces` the `replaceable` module.
```
texastony, line 28: Question: Err... OK... 
so for every Path Syntax OR Langauge-specific feature, create a `replaces` module?

TestModels/ExternV2/README.md
texastony: OK, this question is answered later in this README.
But this sentence is in-accurate, and should be refactored.

A Dafny Developer SHOULD NOT use `extern` if:
1. they will transpile to runtimes that have conflicting package naming requirements
2. the logic will be customized per runtime

For Crypto Tools, (1.) always applies, 
as we transpile to Java and Python, 
which have conflicting package naming requirements.

TestModels/ExternV2/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+This TestModel demonstrates per-path-syntax externs in the `ExternV2Constructor` module and in the `WrappedTestDotNamespaced` module.
+
+A second benefit of the V2 extern system is allowing Dafny to generate code only for particular languages.
+If an extern module requires specific behavior in a particular language, the extern should NOT be replaced in the path-syntax specific module.
+Instead, the developer should write a new file per-language that:
+1. `include`s the path-syntax specific module. (This includes any shared extern definitions in the target language.)
+2. `replaces` the `replaceable` extern module. Here, the developer can add Dafny code specific to the language under generation.
```
texastony, line 36: Question: What should the Language/Runtime specific modules be named?
Should the name follow their package naming scheme? 
Or does it not matter, 
as the `pathy-syntax` specific module will define the name.

TestModels/ExternV2/README.md
texastony, line 1: Issue: It appears we can use 1 stone to solve two problems:
1. I want to name my extern class `foo` in runtimes `x` and `bar` in runtimes `y`
2. I want logic `a` in runtimes `phi` and logic `b` in runtimes `theta`

Since `x` may not equal `phi` or `theta`, we may have lots of `replaces` modules.
While this gives us the most flexibility, 
I wonder if it would not be simpler to remove this flexibility
in favor of "Easy to Use".

Rather than having `replaces` modules for both
the path-syntax and the runtime specific behavior,
could we just establish that every `extern` will have a runtime specific `replaces`
for every runtime we support?

Thus, instead of having:
- for every runtime in runtimes, Index<runtime>.dfy
- for every path-naming in path namings, Index<path-naming>.dfy

We would just have:
- for every runtime in runtimes, Index<runtime>.dfy

I suggest this,
as I find it very unlikely that any runtimes will
share a proper naming convention.

Yes, some runtimes may have naming conventions that 
allow for a package name to be used by both.
But that "common-legal" name is unlikely to be idiomatic in both.

Right now, our .NET packages use Java naming conventions.
Which is legal, but not idiomatic.

It also feels unnecessarily complicated to track 
naming conventions and runtime behavior separately.

TestModels/ExternV2/src/ExternV2Constructor.dfy
```diff
@@ -0,0 +1,26 @@
...
+        // make a static Build method required for extern classes.
+        static method {:extern} Build(input: string)
+        returns (output: Result<ExternV2ConstructorClass, Error>)
+        ensures output.Success?
+        ==>
+            && output.value.Value() == input
+            && fresh(output.value)
+
+        method{:extern} GetValue() returns (output: Result<string, Error>)
+    }
```
texastony, line 25: Question: Will the `replaces` modules change the name of the `ExternV2ConstructorClass`?

TestModels/SharedMakefile.mk
```diff
@@ -98,10 +102,76 @@ dafny-reportgenerator:
...
+        $(COMPILE_SUFFIX_OPTION) \
+        -quantifierSyntax:3 \
+        -unicodeChar:0 \
+        -functionSyntax:3 \
+        -useRuntimeLib \
+        -out $(OUT) \
+        $(SRC_INDEX_TRANSPILE) \
+        -library:$(PROJECT_ROOT)/dafny-dependencies/StandardLibrary/src/Index.dfy \
+        $(patsubst %, -library:$(PROJECT_ROOT)/%/src/Index.dfy, $(LIBRARIES))
+
```
texastony, line 144: Question/Issue/Blocking: `transpile_implementation_v2` is identical to `transpile_implementation`,
with the exception of the source code parameter.

Rather than repeating the whole invocation,
can we make the `src/index.dfy` a parameter,
and invoke `transpile_implementation` with that parameter?

I think you have almost done that, with the if statement on line 127.

texastony (changes requested):
I have left blocking questions.
If it does not make sense to refactor this,
such that we only have `Index<runtime>.dfy`,
than I will approve.

TestModels/ExternV2/src/ExternV2Constructor.dfy
lucasmcdonald3: No. This stays the same in generated code.

NET:

```
      _out0 = simple.dafnyexternv2.internaldafny.ExternV2Constructor.ExternV2ConstructorClass.Build(Wrappers_Compile.Option<Dafny.ISequence<char>>.UnwrapOr((input).dtor_value, Dafny.Sequence<char>.FromString("Error")));
```

Java:

```
    if ((_1_valueOrError0).IsFailure(((dafny.TypeDescriptor<simple.dafnyexternv2.internaldafny.ExternV2Constructor.ExternV2ConstructorClass>)(java.lang.Object)dafny.TypeDescriptor.reference(simple.dafnyexternv2.internaldafny.ExternV2Constructor.ExternV2ConstructorClass.class)), simple.dafnyexternv2.internaldafny.types.Error._typeDescriptor())) {

```

It looks like the only name that is replaced is on the replaceable module.

TestModels/SharedMakefile.mk
lucasmcdonald3: I talk about this in the PR description:

----

I also added new Makefile targets for v2 extern modules. There are pretty useless at the moment. The intention was to use these to run dafny translate, but there are some issues with that:

dafny translate cannot be used to generate tests. Tracking in https://github.com/dafny-lang/dafny/issues/4888
dafny translate would require some work on the Smithy-Dafny side to not generate _Compile suffixes, or to only generate it in some cases. I'm not sure what that would look like right now; this depends on whether we need to use dafny translate to access outer-module.
We will pick these issues up again as part of https://sim.amazon.com/issues/CrypTool-5259 once outer-module is ready, or a replacement for it is ready. I'll leave these in so we don't have to recreate the Makefile structure for them.

----

I kept these in to avoid repeating this work if we have to use `dafny translate`. We can remove it if we prefer though.

TestModels/ExternV2/README.md
lucasmcdonald3: That's inaccurate; we want to use both. This TestModel only implements `replaceable` at the moment, but will implement `outer-module` (or its replacement) once that's ready.

TestModels/ExternV2/README.md
lucasmcdonald3: Something like that. We can use path syntax as long as there are no language-specific features.

I'm open to ideas here. This seems like the least-worst option if we want to avoid repeating extern names.

TestModels/ExternV2/README.md
lucasmcdonald3: Yes, exactly. I called this tradeoff out in the PR description as something reviewers should provide feedback on. We can go in this direction if we think it's better.

TestModels/ExternV2/README.md
lucasmcdonald3: Are the "Language/Runtime specific modules" just the .dfy files?
These can have any name. We just need to point the Makefile to them. Ideally, we might select some default name that the SharedMakefile can use, and the Makefile would override that if needed. i.e. `NetIndex.dfy`, `JavaIndex.dfy`

TestModels/ExternV2/README.md
lucasmcdonald3: Will update

lucasmcdonald3:
Closing in favor of #308 .

We will revive this TestModel once all components of ExternV2 are ready (i.e. `outer-module` or `dfyproject.toml`-stored Module Name).

PR #309, opened 2023-12-19
feat(TestModels): MultipleModels

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

TestModel for multiple Smithy model files in one TestModel.

Also contains Makefile changes to support this structure. These are lifted from https://github.com/aws/aws-cryptographic-material-providers-library-java/blob/main/SharedMakefileV2.mk


TestModels/LanguageSpecificLogic/Model/language-specific-logic.smithy
texastony, line 1: Issue/Question: I can understand why the Makefiles would be executable (` 100644 → 100755 `).
Why would these other files? 
Is this intentional?

TestModels/MultipleModels/README.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+## Development
+1. To add another target runtime support, edit the `Makefile` and add the appropriate recipe to generate the `polymorph` wrappers, and dafny transpilation.
+2. Provide any glue code between dafny and target runtime if required.
+3. Build, execute, and test in the target runtime.
+
+*Example*
+
+`--output-dotnet <PATH>` in the `gradlew run` is used to generate the polymorph wrappers. Similarly `--compileTarget:<RUNTIME>` flags is used in dafny recipe to transpile to C#.
```
texastony: Question: What are these flags doing here?

TestModels/MultipleModels/dafny/DependencyProject/src/SimpleMultiplemodelsDependencyprojectImpl.dfy
```diff
@@ -0,0 +1,21 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+include "../Model/SimpleMultiplemodelsDependencyprojectTypes.dfy"
+
+module SimpleMultiplemodelsDependencyprojectImpl refines AbstractSimpleMultiplemodelsDependencyprojectOperations {
```
texastony, line 5: Nit: This could be hard to fix now, but it the `m` in `model` should be capitalized. 
```suggestion
include "../Model/SimpleMultipleModelsDependencyProjectTypes.dfy"

module SimpleMultipleModelsDependencyProjectImpl refines AbstractSimpleMultipleModelsDependencyProjectOperations {
```

TestModels/SharedMakefile.mk
texastony, line 1: Question: How much of the diff here is novel?
Is any of the diff copied from an already merged SharedMakefile in another repo?

I would like to scope my review to the novel content,
rather than try and reason about all of this at once.

texastony (changes requested):
Two blocking issues:
1. The file mode changes are weird... I do not think Makefiles need to be executable.
I KNOW Markdown and source code files do not. 

2. Answer my question on the SharedMakefile.
If all of this is novel, I am happy to review it.
But if I can compare this to a working, already reviewed, 
SharedMakefile that you copied these changes from,
that would save me some effort.

TestModels/SharedMakefile.mk
lucasmcdonald3: This is a hard question to answer. Almost everything in this file comes from combining existing components in new ways.

I started with many components from the [MPL makefile](https://github.com/aws/aws-cryptographic-material-providers-library-java/blob/main/SharedMakefileV2.mk).
These components center around having multiple models in one project.
I adapted these components to the TestModels, which support projects with `replaceable` modules.  The TestModels' SharedMakefile also handles dependencies differently. It will `polymorph` a project's dependencies, while the MPL's will not.
Combining these required modifying the MPL SharedMakfile targets (e.g. `find ./dafny/**/ ..`, `$(DIR_STRUCTURE_V2)`) to work with `replaceable` modules targets (e.g. `SRC_INDEX_TRANSPILE`).

The only copy-pasted changes from the MPL SharedMakefile are to the `verify` target.

TestModels/MultipleModels/dafny/DependencyProject/src/SimpleMultiplemodelsDependencyprojectImpl.dfy
lucasmcdonald3: I probably won't fix this.

This is actually the correct filename for the namespace.
The namespace is `simple.multiplemodels`.
The capitalization logic is something like "camelcase based on periods", so this becomes "SimpleMultiplemodels".
Fixing this would involve making the namespace `simple.multiple.models`, which is probably more effort than it's worth.

TestModels/MultipleModels/README.md
lucasmcdonald3: cruft... removed

TestModels/LanguageSpecificLogic/Model/language-specific-logic.smithy
lucasmcdonald3: Unintentional. Let me find the right scope of permissions changes.

lucasmcdonald3: Done

lucasmcdonald3: Resolved... not sure what was going on there.

TestModels/SharedMakefile.mk
```diff
@@ -89,6 +93,10 @@ dafny-reportgenerator:
 
 # Dafny helper targets
 
+# Transpile the entire project's impl
+_transpile_implementation_all: TRANSPILE_DEPENDENCIES=$(patsubst %, -library:$(PROJECT_ROOT)/%, $(PROJECT_INDEX))
+_transpile_implementation_all: transpile_implementation
```
texastony: Question: I see this is from the MPL.
Do we ever use it in Smithy-Dafny?

TestModels/SharedMakefile.mk
```diff
@@ -103,28 +111,35 @@ dafny-reportgenerator:
 # and can be the same for all such runtimes.
 # Since such targets are all shared,
 # this is tractable.
-transpile_implementation: SRC_INDEX_TRANSPILE=$(if $(SRC_INDEX),$(SRC_INDEX),src/Index.dfy)
+
+transpile_implementation: SRC_INDEX_TRANSPILE=$(if $(SRC_INDEX),$(SRC_INDEX),src)
```
texastony, line 124: Question: Why is this changed from the MPL version?
Is this an accommodation for the Language Specific Feature?

I suggest adding comment here about what `SRC_INDEX_TRANSPILE` is,
so we can better reason about what this script does.

TestModels/SharedMakefile.mk
```diff
@@ -103,28 +111,35 @@ dafny-reportgenerator:
...
+        -quantifierSyntax:3 \
+        -unicodeChar:0 \
+        -functionSyntax:3 \
+        -useRuntimeLib \
+        -out $(OUT) \
+        -library:$(PROJECT_ROOT)/dafny-dependencies/StandardLibrary/src/Index.dfy \
+        $(TRANSPILE_DEPENDENCIES)
+
+_transpile_test_all: SRC_INDEX_TRANSPILE=$(if $(SRC_INDEX),$(SRC_INDEX),src)
+_transpile_test_all: TEST_INDEX_TRANSPILE=$(if $(TEST_INDEX),$(TEST_INDEX),test)
```
texastony: More comments please.

TestModels/SharedMakefile.mk
```diff
@@ -103,28 +111,35 @@ dafny-reportgenerator:
...
+_transpile_test_all: TRANSPILE_DEPENDENCIES=$(if ${DIR_STRUCTURE_V2}, $(patsubst %, -library:dafny/%/$(SRC_INDEX_TRANSPILE)/Index.dfy, $(PROJECT_SERVICES)), -library:$(SRC_INDEX_TRANSPILE)/Index.dfy)
```
texastony, line 161:
```suggestion
# Transpile the entire project's tests
_transpile_test_all: TRANSPILE_DEPENDENCIES=$(if ${DIR_STRUCTURE_V2}, $(patsubst %, -library:dafny/%/$(SRC_INDEX_TRANSPILE)/Index.dfy, $(PROJECT_SERVICES)), -library:$(SRC_INDEX_TRANSPILE)/Index.dfy)
```

texastony (dismissed):
I am not certain if all the added Makefile logic is used.
But regardless if it is used by Smithy-Dafny,
it might be used by the MPL or another project.

Since we are trying to consolidate,
we should not remove it from here.

Nice work @lucasmcdonald3 !
If you have time to add some comments,
particularly for the content introduced for language-specific,
that would be groovy.

Otherwise, ship it.

TestModels/SharedMakefile.mk
lucasmcdonald3: `transpile_implementation_java` and `transpile_implementation_net` call this target.
It looks like we can consolidate this into `transpile_implementation`, but (as you mentioned) for the sake of SharedMakefile consolidation it might be good to leave as is.

TestModels/SharedMakefile.mk
lucasmcdonald3: Yes, for language-specific logic. Added comment.

texastony (approved):

PR #310, opened 2024-01-24
feat: Compatibility with Dafny 4.4 and upcoming 4.5

robin-aws:
*Description of changes:*

* Add `--properties-file` option to spit out a `project.properties` file in the style of https://github.com/aws/aws-cryptographic-material-providers-library/blob/main/project.properties, and modify all the test module Gradle build files to pick the right Java runtime version using it.
* Address one case of verification regression
* Clean up the CI to remove a duplicative case of testing on multiple Dafny versions
* Refactor WrappersInterop module name to avoid a translation change in 4.4

Manually verified this makes CI pass on the latest Dafny nightly too: https://github.com/smithy-lang/smithy-dafny/actions/runs/7671133054

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
@@ -44,6 +44,7 @@ public class CodegenEngine {
     private final String namespace;
     private final Map<TargetLanguage, Path> targetLangOutputDirs;
     private final DafnyVersion dafnyVersion;
+    private final Optional<Path> propertiesFile;
     // refactor this to only be required if generating Java
     private final AwsSdkVersion javaAwsSdkVersion;
     private final Optional<Path> includeDafnyFile;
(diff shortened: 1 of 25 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
```diff
@@ -117,6 +120,20 @@ public void run() {
                         .formatted(lang.name()));
             }
         }
+
+        if (propertiesFile.isPresent()) {
+            final String propertiesTemplate = IoUtils.readUtf8Resource(
+                    this.getClass(), "/templates/project.properties.template");
```
seebees: Is there a way to have this file not stomp on an existing file?
Or is that just "don't pass the value"?

seebees (dismissed):
LGTM, my guess that that things will come up as people need/want to put more values in this file. Like say other dafny dependencies.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
robin-aws: I was already planning to error if the file already exists :)

robin-aws: Ended up not doing that because all models build the StandardLibrary transitively, so it gets built many times over in CI. :( Since code generation also stomps on existing files I think we should address this in a separate PR.

seebees (approved):
LGTM

PR #312, opened 2024-02-01
fix: Incorrect Dafny version check for Java type descriptors changes

robin-aws:
*Description of changes:*
A previous change accounts for changes in how Dafny uses TypeDescriptors when translating to Java, but a typo makes it apply from Dafny 4.2 and up when it should apply on 4.3 and up.

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
     - name: Populate Dafny versions list
       id: populate-dafny-versions-list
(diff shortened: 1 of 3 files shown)


texastony (approved):
I can't really tell if this works, but I trust you have this figured out.

robin-aws:
> I can't really tell if this works, but I trust you have this figured out.

While I appreciate the trust :) I should have elaborated in the description a bit more. The key point is the CI was set up to test the mechanism thoroughly, but didn't ever test on 4.2. The first commit adds 4.2 and shows the CI breaks for Java pretty badly, but it's fixed when the version threshold typo is fixed.

PR #314, opened 2024-02-06
feat: Add support for automatically patching generated code

robin-aws:
*Description of changes:*

In practice, the projects using `smithy-dafny` such as https://github.com/aws/aws-cryptographic-material-providers-library have had to apply manual edits to generated code, to deal with inconsistencies in AWS SDKs and/or missing `smithy-dafny` features. Besides making it more labour-intensive to regenerate code when necessary, this also makes it challenging to test such projects against different versions of Dafny, especially nightly pre-releases, when code generation has to produce different code for different Dafny versions.

To address this point, this PR adds a new required `--library-root` option. After generating code for a target language, the code generation engine also applies any applicable `<library-root>/codegen-patches/<language>/dafny-<version>.patch` file, selecting the highest such `<version>` that is less than the target `--dafny-version`. This allows the patch files to differ slightly based on the generated code shape. See https://github.com/aws/aws-cryptographic-material-providers-library/pull/195 for examples of using this mechanism.

It also moves the tasks of applying auto-formatters to the generated code into `smithy-dafny` instead of just in the shared makefiles. This makes it much easier to generate these patch files using `git diff`, since the downstream projects affected are already checking in formatted code. `--library-root` is necessary in particular to support formatting .NET code cleanly, since `dotnet format` requires `.csproj` files as arguments. Because the polymorph CLI and the Smithy build plugin have slightly different conventions around library layout, the common CodegenEngine class also now has a `fromSmithyBuildPlugin` boolean to handle both cases, but this is not exposed to users as a CLI option or Smithy build plugin configuration parameter.

Applying `dafny format` also requires that the generated Dafny code can immediately be at least parsed successfully, which implies any dependencies must be already generated. This meant having to tweak orderings in a few Makefiles to satisfy this, but this seemed reasonable given dependencies already needed to be ordered correctly.

Note also that the addition of `--library-root` means that the values of `--output-dafny/java/dotnet/etc` could in the future default to their conventional locations according to the library folder structure that the shared makefiles support.

### DIFF
--- a/TestModels/SharedMakefile.mk
+++ b/TestModels/SharedMakefile.mk
@@ -328,6 +328,10 @@ _polymorph_java: _polymorph_wrapped
 _polymorph_java: POLYMORPH_LANGUAGE_TARGET=java
 _polymorph_java: _polymorph_dependencies
 
+# Dependency for generating Java code
+setup_prettier:
+	npm i --no-save prettier prettier-plugin-java
+
 ########################## .NET targets
 
 transpile_net: | transpile_implementation_net transpile_test_net transpile_dependencies_net
--- a/TestModels/SharedMakefile.mk
+++ b/TestModels/SharedMakefile.mk
@@ -195,6 +195,7 @@ _polymorph:
 	$(GRADLEW) run --args="\
 	$(DAFNY_VERSION_OPTION) \
 	--library-root $(LIBRARY_ROOT) \
+	$(UPDATE_PATCH_FILES_OPTION) \
 	--properties-file $(LIBRARY_ROOT)/project.properties \
 	$(OUTPUT_DAFNY) \
 	$(OUTPUT_DOTNET) \
--- a/TestModels/SharedMakefile.mk
+++ b/TestModels/SharedMakefile.mk
@@ -328,7 +328,7 @@ _polymorph_java: _polymorph_wrapped
 _polymorph_java: POLYMORPH_LANGUAGE_TARGET=java
 _polymorph_java: _polymorph_dependencies
 
-# Dependency for generating Java code
+# Dependency for formatting generating Java code
 setup_prettier:
 	npm i --no-save prettier prettier-plugin-java
 
--- a/TestModels/SharedMakefile.mk
+++ b/TestModels/SharedMakefile.mk
@@ -278,6 +278,8 @@ polymorph_dafny:
 		$(MAKE) _polymorph_dafny || exit 1; \
 	done
 
+_polymorph_dafny: POLYMORPH_LANGUAGE_TARGET=dafny
+_polymorph_dafny: _polymorph_dependencies
 _polymorph_dafny: OUTPUT_DAFNY=\
     --output-dafny $(if $(DIR_STRUCTURE_V2), $(LIBRARY_ROOT)/dafny/$(SERVICE)/Model, $(LIBRARY_ROOT)/Model) \
 	--include-dafny $(STANDARD_LIBRARY_PATH)/src/Index.dfy
@@ -287,8 +289,6 @@ _polymorph_dafny: OUTPUT_DAFNY_WRAPPED=\
     --include-dafny $(STANDARD_LIBRARY_PATH)/src/Index.dfy
 _polymorph_dafny: OUTPUT_LOCAL_SERVICE=--local-service-test
 _polymorph_dafny: _polymorph_wrapped
-_polymorph_dafny: POLYMORPH_LANGUAGE_TARGET=dafny
-_polymorph_dafny: _polymorph_dependencies
 
 # Generates dotnet code for all namespaces in this project
 .PHONY: polymorph_net
--- a/TestModels/SharedMakefile.mk
+++ b/TestModels/SharedMakefile.mk
@@ -243,11 +243,11 @@ _polymorph_dependencies:
 # Generates all target runtime code for all namespaces in this project.
 .PHONY: polymorph_code_gen
 polymorph_code_gen:
-	for service in $(PROJECT_SERVICES) ; do \
+	set -e; for service in $(PROJECT_SERVICES) ; do \
 		export service_deps_var=SERVICE_DEPS_$${service} ; \
 		export namespace_var=SERVICE_NAMESPACE_$${service} ; \
 		export SERVICE=$${service} ; \
-		$(MAKE) _polymorph_code_gen || exit 1; \
+		$(MAKE) _polymorph_code_gen ; \
 	done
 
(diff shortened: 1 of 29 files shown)


TestModels/SharedMakefile.mk
```diff
@@ -326,6 +328,10 @@ _polymorph_java: _polymorph_wrapped
 _polymorph_java: POLYMORPH_LANGUAGE_TARGET=java
 _polymorph_java: _polymorph_dependencies
 
+# Dependency for generating Java code
```
texastony:
```suggestion
# Dependency for formatting generated Java code
```

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/TypeConversionCodegen.java
```diff
@@ -1006,7 +1007,7 @@ protected TokenTree errorFromDanyBody(final TreeSet<StructureShape> errorShapes)
             }
 
             if (dependentNamespaces.size() > 0) {
-                Set<TokenTree> cases = new HashSet<>();
+                List<TokenTree> cases = new ArrayList<>();
```
texastony: Changing this from a Set to List may allow for duplicates.
I think that is currently prevented by `dependentNamespaces` being a Set.
But is there an ordered set we could use instead?

texastony (commented):
This is tested against MPL via https://github.com/aws/aws-cryptographic-material-providers-library/pull/195

How much coverage does https://github.com/aws/aws-cryptographic-material-providers-library/pull/195 give us of this logic?

i.e: Are there patches for .NET, Dafny, & Java? Is more than one Dafny version present?
I am not blocking on this, I just want it recorded what we have covered.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/TypeConversionCodegen.java
robin-aws: Yeah I tried `TreeSet` at first but then realized `TokenTree` wasn't comparable. I can definitely avoid the list.

robin-aws:
> This is tested against MPL via [aws/aws-cryptographic-material-providers-library#195](https://github.com/aws/aws-cryptographic-material-providers-library/pull/195)
> 
> How much coverage does [aws/aws-cryptographic-material-providers-library#195](https://github.com/aws/aws-cryptographic-material-providers-library/pull/195) give us of this logic?
> 
> i.e: Are there patches for .NET, Dafny, & Java? Is more than one Dafny version present? I am not blocking on this, I just want it recorded what we have covered.

This is a good point. The coverage is decent since there are Java and .NET patches, for multiple Dafny versions. But the PR CI for MPL is only going to test on one Dafny version at a time. The nightly build will test on a later version, but will be essentially the same version at times when MPL has just been upgraded.

If it's easy I'll add patch files in a test model somewhere.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/TypeConversionCodegen.java
texastony: We could find a Set that preserves insertion ordering.
Or we can add a comment:
```suggestion
                // dependentNamespace is collision free, so cases will be collision free
                List<TokenTree> cases = new ArrayList<>();
```

Smithy-Dafny has lots of authors modifying logic all over the place.
I am trying to remove or document any logic that assumes the behavior of other logic.

texastony:
> > This is tested against MPL via [aws/aws-cryptographic-material-providers-library#195](https://github.com/aws/aws-cryptographic-material-providers-library/pull/195)
> > How much coverage does [aws/aws-cryptographic-material-providers-library#195](https://github.com/aws/aws-cryptographic-material-providers-library/pull/195) give us of this logic?
> > i.e: Are there patches for .NET, Dafny, & Java? Is more than one Dafny version present? I am not blocking on this, I just want it recorded what we have covered.
> 
> This is a good point. The coverage is decent since there are Java and .NET patches, for multiple Dafny versions. But the PR CI for MPL is only going to test on one Dafny version at a time. The nightly build will test on a later version, but will be essentially the same version at times when MPL has just been upgraded.
> 
> If it's easy I'll add patch files in a test model somewhere.

I am happy with this comment alone.

We know that this works with Java & .NET beneration for multiple Dafny Versions, 
conditioned on one Dafny Version being tested at a time
(which is a very reasonable condition).

texastony (dismissed):
Groovy.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/TypeConversionCodegen.java
robin-aws: For the record the `dependentNamespaces` source is already a `TreeSet`, so the way the code is structured now, duplicates are still impossible. But I changed to use a Stream anyway since it simplified the code too.

texastony (dismissed):

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
```diff
@@ -89,6 +92,12 @@ private static Options getCliOptions() {
             .longOpt("help")
             .desc("print help message")
             .build())
+          .addOption(Option.builder("r")
+            .longOpt("library-root")
+            .desc("root directory of the library")
+            .hasArg()
+            .required()
+            .build())
```
seebees, line 101: Is this `library-root` the root for the whole project, or the language specific project?

seebees (dismissed):
I came to suggest that we have a TestModel that exercises this.

I know we have a lot going on. I am not going to block on it, but... coming back to the test models is difficult. It would be good to get this in.

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
robin-aws: The whole project, same as `LIBRARY_ROOT` from the makefiles. That let me get more mileage out of one option because the patch file directory can also just be convention on top of that root.

seebees (dismissed):
LGTM! Thanks so much for the test!

seebees (dismissed):

seebees (approved):
100% worth :) Thanks!

PR #319, opened 2024-02-22
feat: allow recursion

ajewellamz:
*Description of changes:*

Given a perfectly legal recursive smithy model, smithy-java can find itself in an infinite loop.
Prune DependentShapesWithPaths to avoid this.

Also remove --quatifier-syntax:3, because that blocked me on this change.

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/utils/ModelUtils.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/utils/ModelUtils.java
@@ -310,6 +310,20 @@ public class ModelUtils {
         return outList;
     }
 
+
+    // return a summary of a list, just the first and last elements.
+    public static List<ShapeId> GetSummary(List<ShapeId> list)
+    {
+	if (list.size() <= 2) {
+	    return list;
+	} else {
+	    List<ShapeId> result = new ArrayList<>();
+	    result.add(list.get(0));
+	    result.add(list.get(list.size()-1));
+	    return result;
+	}
+    }
+    
     /**
      * For every ShapeId in {@code initialShapes},
      * with the given {@code model},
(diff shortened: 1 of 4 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/utils/ModelUtils.java
```diff
@@ -326,12 +340,16 @@ public static Set<List<ShapeId>> findAllDependentShapesWithPaths(
...
         Set<List<ShapeId>> pathsToShapes = new LinkedHashSet<>(new LinkedHashSet<>());
+        Set<List<ShapeId>> pathsSummary = new LinkedHashSet<>(new LinkedHashSet<>());
 
         // Breadth-first search via getDependencyShapeIds
         final Queue<List<ShapeId>> toTraverse = new LinkedList<>(initialShapeIdsAsPaths);
         while (!toTraverse.isEmpty()) {
             final List<ShapeId> currentShapeIdWithPath = toTraverse.remove();
-            if (pathsToShapes.add(currentShapeIdWithPath)) {
+
+	    // to avoid cycles, only keep the first list with a given first and last element
```
robin-aws: I may be missing some context, but if the BFS is implemented correctly it shouldn't traverse cycles, and it looks like it has the traversed set already. So why is this necessary?

robin-aws (changes requested):

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/utils/ModelUtils.java
ajewellamz: I can guarantee that cycles are traversed, and you get an arbitrarily large set of arbitrarily large lists.

I'm pretty sure that the problem is with
dependencyShapeIdsWithPaths.forEach(toTraverse::add);
where given (A,B,C) for every dependent X of C we add the list (A,B,C,X) and so we add (A,B,C,B) and (A,B,C,B,C) et cetera.

robin-aws: Ah got it, `toTraverse` is a set of paths rather than nodes, so it doesn't prevent traversing cycles.

Edit: more accurately it's the fact that `pathsToShapes` is a set of paths.

robin-aws: You can make this simpler and clearer by replacing `pathsSummary` by a `Set<ShapeId> visitedShapes` instead. All paths this method produces will always start from the same ShapeID so that's exactly equivalent, and to boot it matches the classic BFS algorithm so it'll be easier to understand.

robin-aws (approved):
Thanks!

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/utils/ModelUtils.java
ajewellamz: Done

texastony (commented):
Comment: We probably need to refactor Smithy-Dafny to use the Knowledge index instead of our various model exploring utility methods.

https://smithy.io/javadoc/1.5.1/software/amazon/smithy/model/knowledge/KnowledgeIndex.html

PR #325, opened 2024-03-02
feat: Provide reusable SmithyDafnyMakefile.mk

robin-aws:
*Description of changes:*

Factors the content of `TestModels/SharedMakefile.mk` to a new top-level `SmithyDafnyMakefile.mk`, which is actually intended to be reused by any `smithy-dafny` project using the CLI through a submodule. Projects will still likely want their own repository-specific `SharedMakefile.mk` in order to set a few variables that are the same for all libraries, which is true for the `TestModels` subfolder here too.

Additional make targets and improvements from https://github.com/aws/aws-cryptographic-material-providers-library and general cleanup are also applied. Highlights:

* Delete `LIBRARIES`, which is unused and was replaced by `PROJECT_DEPENDENCIES`.
* Delete `SMITHY_NAMESPACE`, which is unused and was replaced by `SERVICE_NAMESPACE_<service>`
* Replace `STANDARD_LIBRARY_PATH` with `STD_LIBRARY`.
  * The test models were avoiding the same issue with the standard library trying to build itself by overriding the `transpile_***` targets to not build dependencies. Both mechanisms at once is redundant but safe.
* Introduce `MAX_RESOURCE_COUNT` with a default of `10000000`
* Tweak `GRADLEW` to be the variable to use for building Java runtimes, since the TestModels don't have a `gradlew` in each `runtimes/java` directory, but other projects like the MPL do. Running the smithy-dafny CLI always uses the copy in `codegen`.
* Don't also build `_polymorph_wrapped` by default, just have `TestModels/SharedMakefile.mk` add that instead.
* Invoke `_polymorph_dependencies` before building the current library, since that's necessary in some cases (and was already changed in MPL's makefile)

See https://github.com/smithy-lang/smithy-dafny/pull/326 for a diff of SmithyDafnyMakefile.mk with the old contents of TestModels/SharedMakefile.mk.

### DIFF
--- a/.github/workflows/test_models_net_tests.yml
+++ b/.github/workflows/test_models_net_tests.yml
@@ -112,7 +112,7 @@ jobs:
         working-directory: ./${{ matrix.library }}
         run: |
           make polymorph_dafny DAFNY_VERSION_OPTION="--dafny-version $DAFNY_VERSION"
-          make polymorph_net DAFNY_VERSION_OPTION="--dafny-version $DAFNY_VERSION"
+          make polymorph_dotnet DAFNY_VERSION_OPTION="--dafny-version $DAFNY_VERSION"
 
       - name: Compile ${{ matrix.library }} implementation
         shell: bash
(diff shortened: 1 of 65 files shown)


texastony (dismissed):
See comments on https://github.com/smithy-lang/smithy-dafny/pull/326#pullrequestreview-1915497929

Blocking Comments are:
- Use GRADLEW 
- What is the strip logic?

texastony (approved):
LGTM

PR #326, opened 2024-03-04
DO NOT MERGE - showing diff between TestModels/SharedMakefile.mk and the new SmithyDafnyMakefile.mk 

robin-aws:
See https://github.com/smithy-lang/smithy-dafny/pull/325. This PR copies the contents of SmithyDafnyMakefile.mk over top of TestModels/SharedMakefile.mk, to show what changes as the former is adopted.

### DIFF
--- a/TestModels/SharedMakefile.mk
+++ b/TestModels/SharedMakefile.mk
@@ -1,37 +1,54 @@
 # Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-# These make targets that are shared
-# between all the Dafny projects
-# in this repo.
+# These are make targets that can be shared between all projects
+# that use a common layout.
 # They will only function if executed inside a project directory.
+# See https://github.com/smithy-lang/smithy-dafny/tree/main-1.x/TestModels
+# for examples.
 
-# There are serveral variables that are used here.
+# There are several variables that are used here.
 # The expectation is to define these variables
 # inside each project.
 
 # Variables:
 # MAX_RESOURCE_COUNT -- The Dafny report generator max resource count.
 # 	This is is per project because the verification variability can differ.
-# LIBRARIES -- List of dependencies for the project.
-# 	It should be the list of top leve directory names
-# SMITHY_NAMESPACE -- The smithy namespace to use for code generation. 
-# AWS_SDK_CMD -- the `--aws-sdk` command to generate AWS SDK style interfaces
+# PROJECT_DEPENDENCIES -- List of dependencies for the project.
+# 	It should be the list of top level directory names
+# PROJECT_SERVICES -- List of names of each local service in the project
+# SERVICE_NAMESPACE_<service> -- for each service in PROJECT_SERVICES,
+#   the list of dependencies for that smithy namespace. It should be a list
+#   of Model directories
+# SERVICE_DEPS_<service> -- for each service in PROJECT_SERVICES,
+#   the list of dependencies for that smithy namespace. It should be a list
+#   of Model directories
+# AWS_SDK_CMD -- the `--aws-sdk` command to generate AWS SDK style interfaces.
+# STD_LIBRARY -- path from this file to the StandardLibrary Dafny project.
+# SMITHY_DEPS -- path from this file to smithy dependencies, such as custom traits.
+# GRADLEW -- the gradlew to use when building Java runtimes.
+#   defaults to $(SMITHY_DAFNY_ROOT)/codegen/gradlew
 
-# This evaluates to the local path _of this file_.
-# This means that these are the project roots
-# that are shared by all libraries in this repo.
-PROJECT_ROOT := $(abspath $(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
-# This relative path is for include files between libraries.
-# If an absolut path is used, this path will not be portable.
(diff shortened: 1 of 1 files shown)


TestModels/SharedMakefile.mk
```diff
@@ -162,15 +224,14 @@ transpile_test:
 		-functionSyntax:3 \
 		-useRuntimeLib \
 		-out $(OUT) \
-		-library:$(PROJECT_ROOT)/dafny-dependencies/StandardLibrary/src/Index.dfy \
+		$(if $(strip $(STD_LIBRARY)) , -library:$(PROJECT_ROOT)/$(STD_LIBRARY)/src/Index.dfy, ) \
```
texastony, line 233: Sorry, what is this `if strip` logic?

TestModels/SharedMakefile.mk
```diff
@@ -185,28 +246,31 @@ transpile_dependencies:
 # StandardLibrary is filtered out from dependent-model patsubst list;
 #   Its model is contained in $(LIBRARY_ROOT)/model, not $(LIBRARY_ROOT)/../StandardLibrary/Model.
 _polymorph:
-	@: $(if ${CODEGEN_CLI_ROOT},,$(error You must pass the path CODEGEN_CLI_ROOT: CODEGEN_CLI_ROOT=/path/to/smithy-dafny/codegen/smithy-dafny-codegen-cli));
```
texastony, line 188: I like this option... I will not block on it's removal, 
but I see it's value if I am working on Smithy-Dafny itself,
and want to test it against the MPL/Gazelle with out pushing and pulling it.

TestModels/SharedMakefile.mk
```diff
@@ -185,28 +246,31 @@ transpile_dependencies:
...
 	cd $(CODEGEN_CLI_ROOT); \
-	$(GRADLEW) run --args="\
-	$(DAFNY_VERSION_OPTION) \
+	./../gradlew run --args="\
+	--dafny-version $(DAFNY_VERSION) \
```
texastony, line 257: We have already defined `GRADLEW` above. 
Should we not use it?

TestModels/SharedMakefile.mk
```diff
@@ -185,28 +246,31 @@ transpile_dependencies:
...
-	$(POLYMORPH_OPTIONS)";
+	$(POLYMORPH_OPTIONS) \
+	";
 
 _polymorph_wrapped:
 	@: $(if ${CODEGEN_CLI_ROOT},,$(error You must pass the path CODEGEN_CLI_ROOT: CODEGEN_CLI_ROOT=/path/to/smithy-dafny/codegen/smithy-dafny-codegen-cli));
 	cd $(CODEGEN_CLI_ROOT); \
-	$(GRADLEW) run --args="\
-	$(DAFNY_VERSION_OPTION) \
+	./../gradlew run --args="\
```
texastony, line 278: Again, use GRADLEW

texastony (changes requested):

TestModels/SharedMakefile.mk
robin-aws: I'm not removing it, it just has a default value now, so it's not required but you can override it! We already made this change in the smithy-dafny version: https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/TestModels/SharedMakefile.mk#L33

TestModels/SharedMakefile.mk
robin-aws: The problem this is solving is that the StandardLibrary is automatically considered a dependency of all libraries, but it can't be when building the StandardLibrary itself. 

The TestModels folder's solution is to override lots of targets in `StandardLibrary/Makefile` to do nothing. The MPL, ESDK and DB ESDK use this solution instead: only consider the StandardLibrary a dependency if `STD_LIBRARY` is empty (after stripping whitespace, which I gather is easy to introduce in make variables if you're not careful).

I opted to adopt the latter solution since it aligned a little better with reality and would cause less churn when adopting the SmithyDafnyMakefile.mk in the MPL, ESDK and DB ESDK.

TestModels/SharedMakefile.mk
robin-aws: `GRADLEW` previously meant "what gradlew should we use to run the polymorph CLI", but I'm using it in the unified makefile to mean "what gradlew should we use to build/test/deploy/etc. each `runtimes/java` project.

I did that because the former doesn't really need to be customized AFAICT - it makes sense to always use the `gradlew` sitting next to the polymorph code under `codegen`. But the latter definitely does need be configurable with the current project layouts, because the `TestModels/**/runtimes/java` folders don't have their own gradlew, whereas the projects in the MPL/DB ESDK/ESDK do and are currently building with them.

I should at least add a comment clarifying this to the definition of `GRADLEW`. But you might have a stronger opinion that I'm missing a use case for the current definition of it, or that we should require polymorph libraries to include a copy of gradlew.

texastony: I do not have an opinion.
In theory, it would be optimal for everyone to use the same Gradlew, 
as that might cut down on Gradle Deamon boot ups,
but that theory is pretty shaky.

TestModels/SharedMakefile.mk
texastony: I suggest a comment somewhere.

PR #330, opened 2024-03-28
chore: Cut dafny issues on nightly failures

robin-aws:
*Description of changes:*
Creates an issue that will block the next Dafny release if the nightly build against the latest Dafny prerelease fails.

### DIFF
--- a/.github/workflows/nightly_dafny.yml
+++ b/.github/workflows/nightly_dafny.yml
@@ -27,3 +27,15 @@ jobs:
     uses: ./.github/workflows/test_models_net_tests.yml
     with:
       dafny: 'nightly-latest'
+
+  cut-issue-on-failure:
+    runs-on: ubuntu-20.04
+    needs: [dafny-nightly-verification, dafny-nightly-java, dafny-nightly-net]
+    if: ${{ always() && contains(needs.*.result, 'failure') }}
+    steps:
+      - name: Create release blocker on dafny-lang/dafny
+        run: |
+          gh issue create \
+            --title "Dafny nightly regression from ${{ github.workflow_ref }}" \
+            --label "severity: release-blocker" \
+            --body "See ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly_dafny.yml
```diff
@@ -32,3 +32,18 @@ jobs:
     uses: ./.github/workflows/test_models_rust_tests.yml
     with:
       dafny: 'nightly-latest'
+
+  cut-issue-on-failure:
+    runs-on: ubuntu-latest
+    needs: [dafny-nightly-verification, dafny-nightly-java, dafny-nightly-net, dafny-nightly-rust]
+    if: ${{ contains(needs.*.result, 'failure') }}
```
alex-chew: Is this `if` expression enough to ensure the step executes when the dependencies fail, or is the `always() &&` needed?

.github/workflows/nightly_dafny.yml
```diff
@@ -32,3 +32,18 @@ jobs:
...
+    env:
+      GH_TOKEN: ${{ secrets.CI_TOKEN }}
+    steps:
+      - name: Create release blocker on dafny-lang/dafny
+        run: |
+          gh issue create \
```
alex-chew, line 45: It would be good to create a label for automation-created issues in the Dafny repo, and apply it here with `--label`.

.github/workflows/nightly_dafny.yml
robin-aws: Good catch! I had that on when testing and then removed it thinking for a moment it was only for testing.

.github/workflows/nightly_dafny.yml
robin-aws: That was my original plan, but I found that adding a label to an issue required extra permissions. I figure it's almost as easy for Dafny to search for issues with a prefix rather than a label, and this way the `CI_TOKEN` only needs `public_repo` permissions, which is as minimal/safe as possible.

alex-chew (approved):

PR #331, opened 2024-04-02
chore: Change service constructor result type back (from trait to concrete class)

robin-aws:
*Description of changes:*

Reverts the part of https://github.com/smithy-lang/smithy-dafny/pull/301 that changed the result type of constructing services from the concrete class type to the trait type, which turned out to be breaking in .NET as well as Dafny and was manually reverted in https://github.com/aws/aws-cryptographic-material-providers-library/pull/257. See https://github.com/aws/aws-cryptographic-material-providers-library/pull/267 for the upcoming proper fix to the MPL using this change.

As it turns out this change wasn't necessary in the first place, and the `CreateSuccessOfClient/CreateFailureOfError` helpers are unchanged and still use the trait type, because these helpers are only needed for services that are created by target language code, i.e. SDK-style services and `--local-service-test` wrapped services. Regular local services are constructed in Dafny code so they never need the helper methods.


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
```diff
@@ -2057,21 +2059,16 @@ public TokenTree generateAbstractAwsServiceClass(ServiceShape serviceShape) {
      * See also TestModels/dafny-dependencies/StandardLibrary/src/WrappersInterop.dfy.
      */
     private static TokenTree generateResultOfClientHelperFunctions(String dafnyClientTrait) {
-        final TokenTree createSuccessOfClient = TokenTree
+        return TokenTree
           .of(
-            "// Helper function for the benefit of native code to create a Success(client) without referring to Dafny internals",
+            "// Helper functions for the benefit of native code to create a Success(client) without referring to Dafny internals",
             "function method CreateSuccessOfClient(client: %s): Result<%s, Error> {".formatted(dafnyClientTrait, dafnyClientTrait),
```
texastony, line 2065: Confusion: @robin-aws you suggested I change these to the concrete class for my roll back in the MPL.
Is it OK that they are not that way here?

texastony:
Correction in the overview:
> which turned out to be breaking in Java as well as Dafny

It was actually the MOST breaking in .NET, as that code failed to execute while Java did execute
and even pass all tests and seemed to behave correctly.

texastony (changes requested):
Regrettably, your linked PR to the MPL is insufficient evidence that this works...
You need to pull that MPL into the DB-ESDK and pass tests in Dafny, Java, & .NET to prove that this addresses the breaking change.

Otherwise, I cannot be sure if the changes to the ClientCreation helpers are valid
(from concrete to abstract for `CreateSuccessOf`).

Since I KNOW those were broken during the revert...

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
robin-aws: Yes at the time I thought that was more correct, and then in an earlier version of this change I emitted two versions of each, one for the concrete class and one for the trait. Then I realized I never had use for the concrete class versions so I ended up just not changing the helpers (aside from some formatting cleanup). The PR description spells out why (and I added another sentence).

robin-aws:
> You need to pull that MPL into the DB-ESDK and pass tests in Dafny, Java, & .NET to prove that this addresses the breaking change.

Bring it on. :) The best solution might be to change the Daily CI in the ESDKs to pull in the latest MPL and make sure all CI passes. I could then dry run that against the MPL branch from that PR.

That will be similar (and less complicated) than the nightly CI that does that and regenerates all code to try out the latest Dafny prerelease.

texastony:
Hang on, I can probably do this for you faster... one minute.

texastony (approved):

seebees (commented):
Late to the party, but this does look good. Mostly just updating the models :)

PR #332, opened 2024-04-02
chore: Code generation target for SimpleString test model in Rust

robin-aws:
*Description of changes:*

Manually-crafted code generation target for SimpleString in Rust. Still with some TODOs around error handling, but includes a working set of tests showing what it's like to use the `smithy-rs`-generated interface around the Dafny implementation.

See [this README.md](https://github.com/smithy-lang/smithy-dafny/pull/332/files#diff-54f869ade78eb7b408a2cfab561e68249ef129a76f680a14a4eaa88c1eacc593) for more details. I'm particularly interested in feedback on the `conversions` module, since that's the most novel part compared to all of the boilerplate generated by the existing `smithy-rs` code generator.


TestModels/SimpleTypes/SimpleString/runtimes/rust/src/client.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#[derive(::std::clone::Clone, ::std::fmt::Debug)]
+pub struct Client {
+    handle: ::std::sync::Arc<Handle>,
+}
+
+impl Client {
+    /// Creates a new client from the service [`Config`](crate::Config).
+    #[track_caller]
+    pub fn from_conf(conf: crate::Config) -> Self {
```
ajewellamz: I think this will ave to be
-> Result<Self, Error>

robin-aws: I think so. Interestingly the `smithy-rs` generated version has:

```rust
if let Err(err) = Self::validate_config(&handle) {
    panic!("Invalid client configuration: {err}");
}
```

So invalid config just causes a panic. I expect we at least need to allow constructing SDK wrappers to fail since we could hit arbitrary target language errors? Do we also need this for local services?

ajewellamz: We should never panic! -- only return Results.
For a library "panic!" means "we've lost internal consistency and cannot guarantee the state of your program".

It's the customer's code that should
`if let Err(err) = Self::validate_config(&handle) {`
our libraries should only propagate errors.

TestModels/SimpleTypes/SimpleString/runtimes/rust/src/operation/get_string_single_value.rs
```diff
@@ -0,0 +1,129 @@
...
+        crate::operation::get_string_single_value::GetStringSingleValueError
+    > {
+        let inner_input = crate::conversions::get_string_single_value::_get_string_single_value_input::to_dafny(input);
+        let inner_result = unsafe { (*handle.inner).GetStringSingleValue(&inner_input) };
+        match &*inner_result {
+            crate::implementation_from_dafny::r#_Wrappers_Compile::Result::Success { value } => 
+                Ok(crate::conversions::get_string_single_value::_get_string_single_value_output::from_dafny(value)),
+            crate::implementation_from_dafny::r#_Wrappers_Compile::Result::Failure { error } =>
+                Err(crate::conversions::get_string_single_value::from_dafny_error(error.clone())), // TODO: Why is clone necessary here but not for from_dafny
+            crate::implementation_from_dafny::r#_Wrappers_Compile::Result::_PhantomVariant(_, _) => panic!("Unreachable")
```
ajewellamz: There's some more specific Rust thing for panic!("unreachabe")

ajewellamz: https://doc.rust-lang.org/std/macro.unreachable.html

TestModels/SimpleTypes/SimpleString/runtimes/rust/src/client.rs
robin-aws: Consulted with the author of `smithy-rs` on this. It turns out to be a UX improvement, because failing to construct a client is through to be generally unrecoverable, and not something you would retry or fallback to an alternate configuration. In other words, if the only reasonable thing for callers to do on an `Err` is panic, the client might as well do it for you. 

They did say it's totally reasonable to go the other way, so I'll change to use a Result instead.

TestModels/SimpleTypes/SimpleString/runtimes/rust/README.md
seebees, line 1: Nice. And the link in the root readme :)

TestModels/SimpleTypes/SimpleString/runtimes/rust/src/conversions/get_string_utf8.rs
```diff
@@ -0,0 +1,26 @@
...
+pub fn to_dafny_error(value: crate::operation::get_string_utf8::GetStringUTF8Error) -> ::std::rc::Rc<crate::implementation_from_dafny::r#_simple_dtypes_dsmithystring_dinternaldafny_dtypes::Error> {
+  match value {
+    crate::operation::get_string_utf8::GetStringUTF8Error::Unhandled(unhandled) => 
+      ::std::rc::Rc::new(crate::implementation_from_dafny::r#_simple_dtypes_dsmithystring_dinternaldafny_dtypes::Error::Opaque { obj: Box::into_raw(Box::new(unhandled) as Box<dyn Any>) })
+  }
+}
+
+#[allow(dead_code)]
+pub fn from_dafny_error(dafny_value: ::std::rc::Rc<crate::implementation_from_dafny::r#_simple_dtypes_dsmithystring_dinternaldafny_dtypes::Error>) -> crate::operation::get_string_utf8::GetStringUTF8Error {
+  // TODO: Losing information here, but we have to figure out how to wrap an arbitrary Dafny value as std::error::Error
```
seebees, line 15: Lets add a note about TODO. I'll check the readme

TestModels/SimpleTypes/SimpleString/runtimes/rust/src/operation/get_string/builders.rs
```diff
@@ -0,0 +1,83 @@
...
+        self,
+    ) -> ::std::result::Result<
+        crate::operation::get_string::GetStringOutput,
+        crate::operation::get_string::GetStringError
+    > {
+        let input = self
+            .inner
+            .build()
+            // Using unhandled since GetString doesn't declare any validation,
+            // and smithy-rs seems to not generate a ValidationError case unless there is
```
seebees, line 52: Dos this stuff need a TODO or note so that implementers can find it?

seebees (approved):
It would be nice to get some notes in about TODO and color commentary in the generated code. I thought it was good to have, but could get lost.

PR #336, opened 2024-04-10
feat: add constraint checking for dotnet

ajewellamz:
*Description of changes:*

This ONLY addresses constraint checking for dotnet

pull/257 will address rejecting any types, traits or constraints that we do not or cannot handle properly.

### DIFF
--- a/TestModels/Constraints/Model/Constraints.smithy
+++ b/TestModels/Constraints/Model/Constraints.smithy
@@ -24,7 +24,7 @@ operation GetConstraints {
 
 // These constraints will result
 // a predicate that will define
-// validitaty for a sub set type.
+// validity for a sub set type.
 // The predicate can be tested.
 
 @length(min: 1, max: 10)
@@ -78,11 +78,15 @@ map MapLessThanOrEqualToTen {
   value: String,
 }
 
-@pattern("^[A-Za-z]+$")
-string Alphabetic
+// we don't do patterns yet
+// @pattern("^[A-Za-z]+$")
+// string Alphabetic
 
 @range(min: 1, max: 10)
 integer OneToTen
+
+@range(min: -10, max: 10)
+long TenToTen
 
 @range(min: 1)
 integer GreaterThanOne
@@ -90,20 +94,21 @@ integer GreaterThanOne
 @range(max: 10)
 integer LessThanTen
 
-@uniqueItems
-list MyUniqueList {
-  member: String
-}
+// we don't do uniqueItems yet
+// @uniqueItems
+// list MyUniqueList {
+//   member: String
+// }
 
-@uniqueItems
-list MyComplexUniqueList {
-  member: ComplexListElement
-}
+// @uniqueItems
+// list MyComplexUniqueList {
+//   member: ComplexListElement
+// }
 
-structure ComplexListElement {
-  value: String,
-  blob: Blob,
-}
+// structure ComplexListElement {
+//   value: String,
+//   blob: Blob,
+// }
 
 structure GetConstraintsInput {
   MyString: MyString,
@@ -118,12 +123,13 @@ structure GetConstraintsInput {
   MyMap: MyMap,
   NonEmptyMap: NonEmptyMap,
   MapLessThanOrEqualToTen: MapLessThanOrEqualToTen,
-  Alphabetic: Alphabetic,
+  // Alphabetic: Alphabetic,
   OneToTen: OneToTen,
+  TenToTen: TenToTen,
   GreaterThanOne: GreaterThanOne,
   LessThanTen: LessThanTen,
-  MyUniqueList: MyUniqueList,
-  MyComplexUniqueList: MyComplexUniqueList,
+  // MyUniqueList: MyUniqueList,
+  // MyComplexUniqueList: MyComplexUniqueList,
   MyUtf8Bytes: Utf8Bytes,
   MyListOfUtf8Bytes: ListOfUtf8Bytes,
 }
@@ -141,20 +147,23 @@ structure GetConstraintsOutput {
   MyMap: MyMap,
   NonEmptyMap: NonEmptyMap,
   MapLessThanOrEqualToTen: MapLessThanOrEqualToTen,
-  Alphabetic: Alphabetic,
+  // Alphabetic: Alphabetic,
   OneToTen: OneToTen,
+  TenToTen: TenToTen,
   GreaterThanOne: GreaterThanOne,
   LessThanTen: LessThanTen,
-  MyUniqueList: MyUniqueList,
-  MyComplexUniqueList: MyComplexUniqueList,
+  // MyUniqueList: MyUniqueList,
+  // MyComplexUniqueList: MyComplexUniqueList,
   MyUtf8Bytes: Utf8Bytes,
   MyListOfUtf8Bytes: ListOfUtf8Bytes,
 }
 
 // See Comment in traits.smithy
 @aws.polymorph#dafnyUtf8Bytes
+@length(min: 1, max: 10)
 string Utf8Bytes
 
+@length(min: 1, max: 2)
 list ListOfUtf8Bytes {
   member: Utf8Bytes
 }
--- a/TestModels/Constraints/test/Helpers.dfy
+++ b/TestModels/Constraints/test/Helpers.dfy
@@ -3,160 +3,192 @@
 include "../src/Index.dfy"
 
 module Helpers {
-    import SimpleConstraints
-    import StandardLibrary.UInt
-    import opened SimpleConstraintsTypes
-    import opened Wrappers
+  import SimpleConstraints
+  import opened StandardLibrary.UInt
+  import opened SimpleConstraintsTypes
+  import opened Wrappers
 
-    // UTF-8 encoded "aws-kms"
-    const PROVIDER_ID: UTF8.ValidUTF8Bytes :=
-      var s := [0x61, 0x77, 0x73, 0x2D, 0x6B, 0x6D, 0x73];
-      assert UTF8.ValidUTF8Range(s, 0, 7);
-      s
+  // UTF-8 encoded "aws-kms"
+  const PROVIDER_ID: UTF8.ValidUTF8Bytes :=
+    var s := [0x61, 0x77, 0x73, 0x2D, 0x6B, 0x6D, 0x73];
+    assert UTF8.ValidUTF8Range(s, 0, 7);
+    s
 
-    // Example for overriding GetConstraintInputTemplate with an invalid input
-    // 
-    // method TestGetConstraintWithInvalidMyString(client: ISimpleConstraintsClient)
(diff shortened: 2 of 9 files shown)


TestModels/Constraints/test/Helpers.dfy
seebees, line 1: is this all just formatting?

TestModels/Constraints/Model/Constraints.smithy
```diff
@@ -78,32 +78,37 @@ map MapLessThanOrEqualToTen {
...
 
-@uniqueItems
-list MyUniqueList {
-  member: String
-}
+// we don't do uniqueItems yet
+// @uniqueItems
+// list MyUniqueList {
+//   member: String
+// }
```
seebees, line 101: We don't support them in .Net. We also don't have any code in Java?
Do we add an error if someone tries to do this?

TestModels/Constraints/test/SimpleConstraintsImplTest.dfy
```diff
@@ -34,10 +33,11 @@ module SimpleConstraintsImplTest {
...
-      var input := Helpers.GetConstraintsInputTemplate(overrideToInvalidInput := {"myString"});
+      var input := Helpers.GetValidInput();
+      input := input.(MyString := Some(Helpers.ForceMyString("this string is too long")));
       var ret := client.GetConstraints(input := input);
-      // ret.Failure? in java (good)
-      // ret.Success? in dotnet (bad, but we're working on it)
-      // importantly -- doesn't raise an exception either way
-     }
+      // This client is NOT wrapped
+      // Therefore, it does not do any constraint testing
```
seebees, line 40: I think we want to add this above or something. The point here, is that by not wrapping these things we are relying on Dafny verification. And the fact that this succeeds may imply a soundness bug in the callers code. As opposed to just "well, we don't check it" :)

seebees: Going further, given the extensive testing on the wrapped side, does this test serve any value? would it be better to do nothing here and explain that this is all verified by Dafny and so testing is not useful?

TestModels/Constraints/test/WrappedSimpleConstraintsTest.dfy
```diff
@@ -4,11 +4,565 @@ include "../src/WrappedSimpleConstraintsImpl.dfy"
...
+  }
+
+  // @length(max: 10)
+  method TestGetConstraintWithStringLessThanOrEqualToTen(client: ISimpleConstraintsClient)
+    requires client.ValidState()
+    modifies client.Modifies
+    ensures client.ValidState()
+  {
+    var input := GetValidInput();
+    input := input.(StringLessThanOrEqualToTen := Some(ForceStringLessThanOrEqualToTen("this string is really way too long")));
```
seebees, line 209: Given the conversations about how strings are complicated would it be better to not support this at all?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/TypeConversionCodegen.java
```diff
@@ -251,6 +259,100 @@ public TypeConverter generateBooleanConverter(final BooleanShape booleanShape) {
         return buildConverterFromMethodBodies(booleanShape, fromDafnyBody, toDafnyBody);
     }
 
+    public String generateStringConstraints(final Shape shape, final String value) {
```
seebees, line 262: Again is it better to not support strings at all? Given dafny unicode false support?

TestModels/Constraints/Model/Constraints.smithy
ajewellamz: That's a different topic.
pull/257 is responsible for "we don't support that" errors.

TestModels/Constraints/test/Helpers.dfy
ajewellamz: No, I completely re-did the helpers, into something that actually helped.

TestModels/Constraints/test/SimpleConstraintsImplTest.dfy
ajewellamz: Yes, this test shows that the unwrapped client behaves as expected.
It's not especially interesting, I'll admit, but it seems a reasonable thing to test.

TestModels/Constraints/test/WrappedSimpleConstraintsTest.dfy
ajewellamz: That's a different topic.
pull/257 is responsible for "we don't support that" errors.
This PR makes dotnet behave the same as Java.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/TypeConversionCodegen.java
ajewellamz: Again, that's a different topic.
pull/257 is responsible for "we don't support that" errors.
This PR makes dotnet behave the same as Java.
I suppose that part pf pull/257 might be to rip these out, since the smithy file will have to remove any length constraints on strings.

TestModels/Constraints/test/WrappedSimpleConstraintsTest.dfy
```diff
@@ -4,11 +4,565 @@ include "../src/WrappedSimpleConstraintsImpl.dfy"
...
+  // @range(min: 1, max: 10)
+  method TestGetConstraintWithOneToTen(client: ISimpleConstraintsClient)
+    requires client.ValidState()
+    modifies client.Modifies
+    ensures client.ValidState()
+  {
+    var input := GetValidInput();
+    input := input.(OneToTen := Some(ForceOneToTen(1000)));
+    var ret := client.GetConstraints(input := input);
+    expect ret.Failure?;
```
texastony, line 83: Nit/Issue: I keep running into tests where we just `expect ret.Failure?` and do not validate what caused the failure or what the failure was.

These tests sometimes mask bugs.
As an example, there is a test that asserts that if no Branch Key ID exists,
the Keystore's Get* methods for that Branch Key ID will fail.

And the Keystore does fail! 
But with a bizarre KMS Exception that no customer could use to realize they were asking for a non-existance Branch Key ID.

Is it possible to bulk replace `ret.Failure?` here with something like `ret.Failure?` AND `ret.Failure == IllegalInputException`?

ajewellamz: I agree completely, but unfortunately, the error is an Opaque error that came from an exception, and there's no way (as far as I know) to get the actual exception type or the error text out of an Opaque error.
Further, things will likely be different between Java and Dotnet, even if we could get at that stuff; since the errors are generated from custom native code.

TestModels/Constraints/test/Helpers.dfy
```diff
@@ -3,160 +3,192 @@
...
+  function method ForceListOfUtf8Bytes(value : seq<Utf8Bytes>) : ListOfUtf8Bytes
+  {
+    assume {:axiom} IsValid_ListOfUtf8Bytes(value);
+    var myListOfUtf8Bytes: ListOfUtf8Bytes := value;
+    myListOfUtf8Bytes
+  }
+
+  function method ForceLessThanTen(value : int) : LessThanTen
+  {
+    assume {:axiom} ValidInt32(value);
```
lucasmcdonald3, line 77: We might want to loop in @robin-aws.
He pushed back on testing failure cases for SimpleConstraints using `assume` statements.
(I am on board with what you've done)

TestModels/Constraints/test/WrappedSimpleConstraintsTest.dfy
texastony: I just logged in to write the same thing. 
Which is a pretty severe limitation of this Test Model approach we have endorsed. 

@seebees And @robin-aws  for awareness. 

Not that we are going to take any action to resolve this, but we should bear in mind that our limited error expressions mixed with our wrapping test strategy ensures we can never assert the exception type.

texastony (approved):

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/TypeConversionCodegen.java
```diff
@@ -260,20 +362,23 @@ public TypeConverter generateStringConverter(final StringShape stringShape) {
...
-        final TokenTree toDafnyBody = Token.of("return Dafny.Sequence<char>.FromString(value);");
+        final TokenTree toDafnyBody = Token.of(constraints + "return Dafny.Sequence<char>.FromString(value);");
         return buildConverterFromMethodBodies(stringShape, fromDafnyBody, toDafnyBody);
     }
 
     public TypeConverter generateIntegerConverter(final IntegerShape integerShape) {
+        final String constraints = generateNumberConstraints(integerShape, "value");
         final TokenTree fromDafnyBody = Token.of("return value;");
-        final TokenTree toDafnyBody = Token.of("return value;");
+        final TokenTree toDafnyBody = Token.of(constraints + "return value;");
```
texastony, line 374: Blocking: Are we not putting these into the `validate` method?
It appears we pushing these `constraint` checks into the Type Conversion Methods,
which definitely would solve the problem.

But I thought the plan was to put them in structure's `validate` method,
and then ensure `validate` is always called in the `toDafnyBody` methods.

texastony (changes requested):

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/TypeConversionCodegen.java
texastony: @ajewellamz and @texastony chatted on a call.
My dreams of a useful and complete `Validate` method are punted to another day.
But @ajewellamz in this PR or a very fast follow will ensure that `Validate` is called in the to Dafny methods for all structures. 

I think I am going to leave this PR blocked until that is done,
but if @ajewellamz tells me that he really wants to do it another PR, 
I will yield.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/TypeConversionCodegen.java
```diff
@@ -451,7 +558,12 @@ private TypeConverter generateRegularStructureConverter(final StructureShape str
                 constructorArgs.parenthesized(),
                 Token.of(';')
         );
+        String validate = "value.Validate();";
+        if (AwsSdkNameResolverHelpers.isInAwsSdkNamespace(structureShape.getId())) {
+            validate = "";
+        }
```
texastony, line 564: Groovy.

texastony (approved):
Most Groovy.
Thank you @ajewellamz

PR #340, opened 2024-04-15
chore: fix for constraints with weird names

ajewellamz:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/ServiceCodegen.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/ServiceCodegen.java
@@ -4,10 +4,20 @@
 package software.amazon.polymorph.smithydotnet;
 
 import com.google.common.annotations.VisibleForTesting;
+import java.math.BigDecimal;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Stream;
 import software.amazon.polymorph.smithydotnet.nativeWrapper.NativeWrapperCodegen;
 import software.amazon.polymorph.traits.ExtendableTrait;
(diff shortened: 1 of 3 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/ServiceCodegen.java
```diff
@@ -689,6 +690,16 @@ private String getConstraints(
     };
   }
 
+  public String capitalizeSegment(final String segment) {
+    int index = segment.indexOf('_');
+    if (index < 0) {
+      return nameResolver.capitalizeNamespaceSegment(segment);
+    } else {
+      System.err.println(StringUtils.capitalize(segment));
```
robin-aws:
```suggestion
```

Assuming this was debugging

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/ServiceCodegen.java
```diff
@@ -689,6 +690,16 @@ private String getConstraints(
     };
   }
 
+  public String capitalizeSegment(final String segment) {
```
robin-aws: Given underscores can appear in any Smithy namespaces/identifiers by the looks of it (https://smithy.io/2.0/spec/model.html#shape-id-abnf), we might hit this same error in several other places in the future. It would be a lot better to push this fix into `capitalizeNamespaceSegment` instead, and to split on `_` and handle each piece the same, instead of assuming we should give up and use `StringUtils.capitalize` instead.

robin-aws (changes requested):

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/ServiceCodegen.java
ajewellamz: Done

texastony (approved):

robin-aws (approved):
MUCH nicer, thanks!

PR #341, opened 2024-04-15
feat: support enum types in Smithy 2.0 

kessplas:
*Issue #, if available:*

*Description of changes:*

This is an updated version of https://github.com/smithy-lang/smithy-dafny/pull/202. That branch is very far behind `main-1.x` particularly after the recent Java formatting changes. To avoid merge conflict hell, I'm submitting a new PR from a new branch. 

Currently, only .NET is supported (via TestModels). I haven't attempted Java. 


.github/workflows/test_models_net_tests.yml
robin-aws, line 1: Flagging for myself that this difference (`@enum` being supported for only .NET so far) would be a great thing to explicitly flag in #257 :)

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
```diff
@@ -87,8 +87,16 @@ public Map<Path, TokenTree> generate() {
           // Sort by shape ID for deterministic generated code
           .collect(Collectors.toCollection(TreeSet::new))
           .stream()
-          .map(this::generateCodeForShape)
-          .flatMap(Optional::stream)
+          .flatMap(shape -> {
+            final Optional<TokenTree> tokens = generateCodeForShape(shape);
+            if (tokens.isEmpty()) {
+              LOGGER.info(
```
robin-aws, line 93: @alex-chew Intrigued by this - I feel like either it should be a warning, if we think this points to a degenerate model or likely internal bug or something like that, or should not be output at all, if sometimes it's perfectly reasonable to emit no code for some shapes, and outputting this might cause users to wonder if it's a problem.

My inclination is towards the latter. We could always add a -v verbose mode that outputs these kinds of messages for debugging.

robin-aws (approved):
LGTM based on prior reviews on https://github.com/smithy-lang/smithy-dafny/pull/202, just a non-blocking comment for something I didn't pick up on before.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
alex-chew: Yep, it seems reasonable to only output this in a verbose mode. I think I probably originally included this to aid in debugging.

alex-chew (approved):

PR #343, opened 2024-04-19
chore: address lints in SimpleString Rust

andrewbanchich:
*Description of changes:*

- add `#![allow(deprecated)]` to silence warning for matching on deprecated `GetStringUTF8Error::Unhandled` enum variant
- remove unused `use core::panic;` import
- format with `rustfmt`


TestModels/SimpleTypes/SimpleString/runtimes/rust/dafny_impl/Cargo.lock
robin-aws, line 1: AFAICT the best practice is usually to commit `Cargo.lock` for binaries, but not for libraries, since the actual dependency resolution happens in consuming projects instead. So I think we should not commit this file.

robin-aws (changes requested):
Great first start! Just one change to make.

TestModels/SimpleTypes/SimpleString/runtimes/rust/dafny_impl/Cargo.lock
andrewbanchich: updated!

robin-aws (approved):
🚀

Issue #298, opened 2023-10-11
Use @smithy.test#smokeTests in test models (was: Add tests to verify how AWS SDKs handle optional types)


seebees:
In the .NET SDK an optional value is stored privately as `null`.
There is also a private `IsSet` method that will identify if the value is set or not.

The .NET SDK will return never return `null`.
If you create a DDB input with `Limit` for example, but do not set the `Limit` value,
the underlying value will be `null`.
However, if you try and get this value, you will not get `null` you will get `0`.

We should add a test model to check this.
Since it is likely to come up in other languages.

texastony:
Internally, 
we tracked this as [CrypTool-5098](https://sim.amazon.com/issues/CrypTool-5098),
which I will now resolve in favor of this Issue.

A similar edge case was addressed in Java via https://github.com/smithy-lang/smithy-dafny/pull/191.

Additionally, 
any Test Model we add to address this
will require Native (or Extern) code.

The request MUST originate on the Native side,
and then MUST be passed to Dafny, 
for us to assert that the ToDafny conversion
respects unset values.

robin-aws:
A great trait for this was just added to the Smithy specification: https://smithy.io/2.0/additional-specs/smoke-tests.html#smoketests-trait

The trait defines simple requests to send to a service and an assertion that the operation either succeeds or fails. This is fairly ideal for testing constraints such as `@required` in test models:

```smithy
@aws.polymorph#localService(
  sdkId: "SimpleConstraints",
  config: SimpleConstraintsConfig,
)
service SimpleConstraints {
  version: "2021-11-01",
  resources: [],
  operations: [ GetConstraints ],
  errors: [ SimpleConstraintsException ],
}

@smithy.test#smokeTests(
    [
        {
            id: "PresentSuccess"
            params: {mustBePresent: "No problem"}
            expect: {
                success: {}
            }
        }
        {
            id: "AbsenceError"
            params: {}
            expect: {
                failure: {}
            }
        }
    ]
)
operation GetConstraints {
  input: GetConstraintsInput,
  output: GetConstraintsOutput,
}

structure GetConstraintsInput {
  @required mustBePresent: String
}

...

```

The key difference between this testing method and Dafny-based tests: the `params` document can specify invalid data, which Dafny code can't given that we embed constraint traits into verified preconditions/non-optional types/subset types/etc.

[...]

PR #344, opened 2024-04-22
feat: add SimpleBoolean TestModel for Rust

andrewbanchich:
*Description of changes:*

adds Rust code expected to be generated by Dafny -> Rust for `SimpleBoolean`


TestModels/SimpleTypes/SimpleBoolean/runtimes/rust/tests/simple_boolean_test.rs
```diff
@@ -0,0 +1,49 @@
...
+      TestGetBooleanSingleValue(client);
+  }
+*/
+
+/*method TestGetBoolean(client: ISimpleTypesBooleanClient)
+{
+    var ret :- expect client.GetBoolean(SimpleBoolean.Types.GetBooleanInput(value:= Some("TEST_SIMPLE_BOOLEAN_VALUE")));
+    expect ret.value.UnwrapOr("") == "TEST_SIMPLE_BOOLEAN_VALUE";
+    print ret;
+} */
```
andrewbanchich: still need to update these doc comments and the tests. working on that now. just posting PR to get initial feedback.

robin-aws: FYI I'm working on https://github.com/smithy-lang/smithy-dafny/issues/298 right now, which will provide a standard way to generate some of these target language tests.

TestModels/SimpleTypes/SimpleBoolean/runtimes/rust/Cargo.lock
robin-aws, line 1: Same comment as on https://github.com/smithy-lang/smithy-dafny/pull/343 about not checking lock files in

robin-aws:
Having a look now, but one important thing to fix right away: make sure you uncomment this line so that the tests are actually run in CI: https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/.github/workflows/test_models_rust_tests.yml#L36

TestModels/SimpleTypes/SimpleBoolean/runtimes/rust/dafny_impl/src/implementation_from_dafny.rs
```diff
@@ -0,0 +1,501 @@
...
+    pub struct WrappersInterop {}
+
+    impl WrappersInterop {
+        pub fn new() -> Self {
+            WrappersInterop {}
+        }
+        pub fn _allocated() -> *mut Self {
+            ::dafny_runtime::allocate::<Self>()
+        }
+        pub fn CreateBooleanSome(
```
robin-aws, line 490: I totally get why you removed the `CreateStringSome/None` functions, but they are actually coming from here: https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/TestModels/dafny-dependencies/StandardLibrary/src/WrappersInterop.dfy So they will always be emitted by the Dafny compiler not matter what Smithy model we're using, and should still be here.

(I only spotted this by diffing this file against the SimpleString version. :) )

robin-aws: I also notice this file has `rustfmt` applied while I think the SimpleString version doesn't, but I think that's actually better since we expect `dafny` to apply `rustfmt` itself anyway, so we should fix the SimpleString version at some point instead.

robin-aws (changes requested):
Looks great! Just a couple of minor fixups.

TestModels/SimpleTypes/SimpleBoolean/runtimes/rust/Cargo.lock
andrewbanchich: fixed

TestModels/SimpleTypes/SimpleBoolean/runtimes/rust/dafny_impl/src/implementation_from_dafny.rs
andrewbanchich: ah, good catch! i've updated the PR to add those back.

robin-aws (approved):

PR #353, opened 2024-05-03
feat: add SimpleBlob TestModel for Rust

andrewbanchich:
*Description of changes:*

adds `SimpleBlob` test model


TestModels/SimpleTypes/SimpleBlob/runtimes/rust/dafny_impl/src/implementation_from_dafny.rs
```diff
@@ -0,0 +1,546 @@
...
+    /*
+     datatype GetBlobInput = | GetBlobInput (
+      nameonly value: Option<blob> := Option.None
+    )
+    */
+    #[derive(Clone)]
+    pub enum GetBlobInput {
+        GetBlobInput {
+            value: ::std::rc::Rc<
+                super::r#_Wrappers_Compile::Option<::std::vec::Vec<::std::primitive::u8>>,
```
MikaelMayer: Since the Dafny type here is `Option<seq<uint8>>`, the generated type will be
```suggestion
                super::r#_Wrappers_Compile::Option<::dafny_runtime::Sequence<::std::primitive::u8>>,
```

TestModels/SimpleTypes/SimpleBlob/runtimes/rust/src/operation/get_blob/_get_blob_input.rs
```diff
@@ -0,0 +1,64 @@
+// Code generated by software.amazon.smithy.rust.codegen.smithy-rs. DO NOT EDIT.
+#[allow(missing_docs)] // documentation missing in model
+#[non_exhaustive]
+#[derive(::std::clone::Clone, ::std::cmp::PartialEq, ::std::fmt::Debug)]
+pub struct GetBlobInput {
+    #[allow(missing_docs)] // documentation missing in model
+    pub value: ::std::option::Option<::dafny_runtime::Sequence<::std::primitive::u8>>,
```
robin-aws: Ah this should be the native idiomatic Rust type, not the Dafny type, since it's the struct that customers will work with.

TestModels/SimpleTypes/SimpleBlob/runtimes/rust/tests/simple_blob_test.rs
```diff
@@ -0,0 +1,81 @@
...
+       // i.e. GetBlob(GetBlobInput(seq)) == GetBlobInput(seq)
+       expect ret.value.UnwrapOr([0x0]) == convertedBlobInput.value.value;
+       // From above, GetBlobInput(seq) value == seq, so the below should be redundant
+       expect ret.value.UnwrapOr([0x0]) == s;
+       print ret;
+   }
+*/
+#[tokio::test]
+async fn test_get_blob() {
+    let s = ::dafny_runtime::Sequence::from_array(&vec![0x0, 0x1, 0x2]);
```
robin-aws: This is the primary clue the types aren't right, because these tests shouldn't refer to any Dafny types.

robin-aws (changes requested):

TestModels/SimpleTypes/SimpleBlob/runtimes/rust/src/operation/get_blob/_get_blob_input.rs
```diff
@@ -0,0 +1,60 @@
+// Code generated by software.amazon.smithy.rust.codegen.smithy-rs. DO NOT EDIT.
+#[allow(missing_docs)] // documentation missing in model
+#[non_exhaustive]
+#[derive(::std::clone::Clone, ::std::cmp::PartialEq, ::std::fmt::Debug)]
+pub struct GetBlobInput {
+    #[allow(missing_docs)] // documentation missing in model
+    pub value: ::std::option::Option<::std::vec::Vec<::std::primitive::u8>>,
```
robin-aws, line 7: Just FYI, I see that in practice `smithy-rs` uses `::aws_smithy_types::Blob` instead: https://github.com/smithy-lang/smithy-rs/blob/main/rust-runtime/aws-smithy-types/src/blob.rs

We do want to avoid AWS-specific types so I think using `Vec[u8]` is the right call, but I mention this in case there are compatibility concerns down the road.

TestModels/SimpleTypes/SimpleBlob/runtimes/rust/src/conversions/get_blob/_get_blob_input.rs
```diff
@@ -26,9 +26,10 @@ pub fn from_dafny(
         dafny_value.value().as_ref(),
         ::simple_blob_dafny::_Wrappers_Compile::Option::Some { .. }
     ) {
-        Some(::std::rc::Rc::unwrap_or_clone(
-            dafny_value.value().Extract().to_array(),
-        ))
+        Some(
+            ::std::rc::Rc::try_unwrap(dafny_value.value().Extract().to_array())
```
robin-aws, line 30: This looks like the right approach AFAICT. I'm not 100% sure it's the best we can do in the limit, but I am sure that it's much easier to optimize this kind of thing once we're actually generating the code, so it should be very quick to fix if we do realize we can do better latter. E.g. we may want to add this conversion to the Dafny Rust runtime.

robin-aws (dismissed):

MikaelMayer:
Please import the generated file [`runtimes/rust/dafny_impl/src/implementation_from_dafny.rs`](https://github.com/smithy-lang/smithy-dafny/blob/36d2089a7bd5e6290f9c28f80183f27252fd938e/TestModels/SimpleTypes/SimpleBlob/runtimes/rust/dafny_impl/src/implementation_from_dafny.rs) of PR https://github.com/smithy-lang/smithy-dafny/pull/368 so that there will be no merge conflicts.

and I can help you generate the future files after #368 (need to install Dafny locally, on the branch feat-rust, and then run `DAFNY=path/to/Dafny.exe make dafny_benerate` in a `SimpleBlob`-like folder

robin-aws (approved):

PR #358, opened 2024-05-13
chore: update github CI rust toolchain to 1.78.0

andrewbanchich:
*Issue #, if available:*

*Description of changes:*


robin-aws:
Ideally we shouldn't require a newer version of the toolchain than the AWS SDKs: https://github.com/awslabs/aws-sdk-rust?tab=readme-ov-file#supported-rust-versions-msrv

Would `1.75.0` be good enough?

andrewbanchich:
looks like [that was introduced in 1.76](https://doc.rust-lang.org/std/rc/struct.Rc.html#method.unwrap_or_clone). i'll update with an alternative approach.

PR #379, opened 2024-05-21
ci: Use Dafny feature branch for testing Rust support

robin-aws:
*Description of changes:*

Moves the CI to use the [feat-rust branch of Dafny](https://github.com/dafny-lang/dafny/tree/feat-rust) instead of a specific nightly prerelease, since the branch should be a stable target as we work on merging that work to master.

Note that the patch files don't change much with this switch because the `src_for_rust` subset isn't changing yet - I'm having trouble adding more files there for some reason, and will follow up in a different PR to improve that.

### DIFF
--- a/.github/actions/build_dafny_from_source/action.yml
+++ b/.github/actions/build_dafny_from_source/action.yml
@@ -23,6 +23,7 @@ runs:
         ref: ${{ inputs.dafny }}
 
     - name: Load Z3
+      shell: bash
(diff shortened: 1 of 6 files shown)


robin-aws:
Merging this into https://github.com/smithy-lang/smithy-dafny/pull/368 and will make that actually update the build process and patch files to take advantage of the branch.

PR #385, opened 2024-05-23
feat: Pass -rlimit $MAX_RESOURCE_COUNT to Dafny instead of the report generator, don't -trace by default

robin-aws:
*Description of changes:*

The Dafny team has found that passing a direct `-rlimit` when verifying is better than using the report generator to enforce it after the fact, because individual expensive assertion batches can be given a higher limit using `{:rlimit N}`, allowing exceptions while still enforcing a lower limit in the rest of the code base. See the new comment on `MAX_RESOURCE_COUNT`:

```
# MAX_RESOURCE_COUNT -- The Dafny verification max resource count.
#   This is is per project because the verification variability can differ.
#   Individual functions/methods/lemmas can override this limit using `{:rlimit N}`,
#   but be aware that the attribute multiplies N by 1000!
#   N can be 0 to enforce no limit, but this is not recommended.
#   You can also temporarily override or disable this limit during development
#   with commands like `make verify MAX_RESOURCE_COUNT=0`.
```

Also made `make verify` not pass `-trace` by default, since this creates very noisy output in CI and makes it harder to locate failures. To add the option back, use `make verify DAFNY_OPTIONS=-trace`.

### DIFF
--- a/SmithyDafnyMakefile.mk
+++ b/SmithyDafnyMakefile.mk
@@ -12,8 +12,10 @@
 # inside each project.
 
 # Variables:
-# MAX_RESOURCE_COUNT -- The Dafny report generator max resource count.
+# MAX_RESOURCE_COUNT -- The Dafny verification max resource count.
 # 	This is is per project because the verification variability can differ.
+#   Individual functions/methods/lemmas can override this limit using `{:rlimit N}`,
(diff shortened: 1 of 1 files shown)


seebees (approved):
Looks good. Having `{rlimit}` for individual locations is so much better!

PR #387, opened 2024-05-24
feat: SimpleDouble for Rust

ajewellamz:
*Issue #, if available:*

*Description of changes:*


TestModels/SimpleTypes/SimpleDouble/codegen-patches/rust/dafny-4.5.0.patch
```diff
@@ -0,0 +1,1749 @@
...
++    
+ pub mod r#_simple_dtypes_dsmithydouble_dinternaldafny_dtypes {
+     #[derive(PartialEq, Clone)]
+     pub enum DafnyCallEvent<I: ::dafny_runtime::DafnyType, O: ::dafny_runtime::DafnyType> {
+@@ -103,15 +103,12 @@ pub mod r#_simple_dtypes_dsmithydouble_dinternaldafny_dtypes {
+     #[derive(PartialEq, Clone)]
+     pub enum GetDoubleInput {
+         GetDoubleInput {
+-            value: ::std::rc::Rc<super::r#_Wrappers_Compile::Option<::dafny_runtime::Sequence<u8>>>,
++            value: ::std::rc::Rc<super::r#_Wrappers_Compile::Option<f64>>,
```
robin-aws: Ah no the original is actually correct for Dafny. That's because since Dafny has no native support for floating point values (`real` is actually closer to rational numbers at runtime). Dafny's current representation of a Smithy Double is the raw 8 bytes as a `seq<uint8>`. That means the conversion code needs to call `f64::from_be_bytes` to convert.

robin-aws (changes requested):

TestModels/SimpleTypes/SimpleDouble/codegen-patches/rust/dafny-4.5.0.patch
```diff
@@ -0,0 +1,1762 @@
...
+     pub enum DafnyCallEvent<I: ::dafny_runtime::DafnyType, O: ::dafny_runtime::DafnyType> {
+@@ -108,10 +108,7 @@ pub mod r#_simple_dtypes_dsmithydouble_dinternaldafny_dtypes {
+     }
+ 
+     impl GetDoubleInput {
+-        pub fn value(
+-            &self,
+-        ) -> &::std::rc::Rc<super::r#_Wrappers_Compile::Option<::dafny_runtime::Sequence<u8>>>
+-        {
++        pub fn value(&self) -> &::std::rc::Rc<super::r#_Wrappers_Compile::Option<::dafny_runtime::Sequence<u8>>> {
```
robin-aws: Changing the formatting is unnecessary but harmless, will be cleaned up once we stop patching in general later :)

TestModels/SimpleTypes/SimpleDouble/codegen-patches/rust/dafny-4.5.0.patch
```diff
@@ -0,0 +1,1762 @@
...
+             match self {
+                 GetDoubleInput::GetDoubleInput { value } => value,
+             }
+@@ -142,7 +139,7 @@ pub mod r#_simple_dtypes_dsmithydouble_dinternaldafny_dtypes {
+     }
+ 
+     impl Eq for GetDoubleInput {}
+-
++/*
```
robin-aws: What's broken about the code you're commenting out here?

TestModels/SimpleTypes/SimpleDouble/codegen-patches/rust/dafny-4.5.0.patch
```diff
@@ -0,0 +1,1762 @@
...
++        >;
+     }
+ 
+     #[derive(PartialEq, Clone)]
+@@ -410,20 +417,15 @@ pub mod r#_simple_dtypes_dsmithydouble_dinternaldafny_dtypes {
+     pub type OpaqueError =
+         ::std::rc::Rc<super::r#_simple_dtypes_dsmithydouble_dinternaldafny_dtypes::Error>;
+ }
+-pub mod r#_SimpleSmithyDoubleOperations_Compile {
++pub mod r#_SimpleDoubleImpl_Compile {
```
robin-aws: Shouldn't be changed, smithy namespace is `simple.types.smithyDouble`

TestModels/SimpleTypes/SimpleDouble/codegen-patches/rust/dafny-4.5.0.patch
```diff
@@ -0,0 +1,1762 @@
...
+-
++*/
+     impl ::std::default::Default for GetDoubleOutput {
+         fn default() -> GetDoubleOutput {
+             GetDoubleOutput::GetDoubleOutput {
+@@ -302,6 +296,19 @@ pub mod r#_simple_dtypes_dsmithydouble_dinternaldafny_dtypes {
+                 ::std::rc::Rc<super::r#_simple_dtypes_dsmithydouble_dinternaldafny_dtypes::Error>,
+             >,
+         >;
++        fn GetDoubleKnownValueTest(
```
robin-aws: This shouldn't be here since the smithy model doesn't actually have a `GetDoubleKnownValueTest` yet

TestModels/SimpleTypes/SimpleDouble/codegen-patches/rust/dafny-4.5.0.patch
```diff
@@ -0,0 +1,1762 @@
...
+ }
+-pub mod r#_SimpleSmithyDoubleOperations_Compile {
++pub mod r#_SimpleDoubleImpl_Compile {
+     pub struct _default {}
+ 
+     impl _default {
+         pub fn _allocate_rcmut() -> ::dafny_runtime::Object<Self> {
+             ::dafny_runtime::allocate_rcmut::<Self>()
+         }
+-        pub fn r#_ValidInternalConfig_q(
```
robin-aws: Why delete this?

TestModels/SimpleTypes/SimpleDouble/runtimes/rust/src/operation/get_double/_get_double_input.rs
```diff
@@ -0,0 +1,60 @@
+// Code generated by software.amazon.smithy.rust.codegen.smithy-rs. DO NOT EDIT.
+#[allow(missing_docs)] // documentation missing in model
+#[non_exhaustive]
+#[derive(::std::clone::Clone, ::std::cmp::PartialEq, ::std::fmt::Debug)]
+pub struct GetDoubleInput {
+    #[allow(missing_docs)] // documentation missing in model
+    pub value: ::std::option::Option<::std::vec::Vec<::std::primitive::u8>>,
```
robin-aws: Ah sorry I wasn't clearer, this should still be `f64`, cause that's the idiomatic Rust type for a Smithy long.

But since the Dafny-generated type is `dafny_runtime::Sequence<u8>`, the conversion `from_dafny` function needs to call something like `f64.from_be_bytes(...)` on the sequence of bytes.

robin-aws (changes requested):

TestModels/SimpleTypes/SimpleDouble/runtimes/rust/src/conversions/get_double/_get_double_input.rs
```diff
@@ -0,0 +1,43 @@
...
+    dafny_value: ::std::rc::Rc<
+        ::simple_double_dafny::r#_simple_dtypes_dsmithydouble_dinternaldafny_dtypes::GetDoubleInput,
+    >,
+) -> crate::operation::get_double::GetDoubleInput {
+    let value = if matches!(
+        dafny_value.value().as_ref(),
+        ::simple_double_dafny::_Wrappers_Compile::Option::Some { .. }
+    ) {
+        let my_rc_vec : ::std::rc::Rc<Vec<u8>> = dafny_value.value().Extract().to_array();
+        let my_vec : Vec<u8> = (*my_rc_vec).clone();
```
robin-aws: This part can be replaced by a call to `::dafny_runtime::dafny_runtime_conversions::dafny_sequence_to_vec` at least.

robin-aws (changes requested):
Almost there, the only issue is that when I said...

> This shouldn't be here since the smithy model doesn't actually have a GetDoubleKnownValueTest yet

I failed to explicitly specify that this applied to the whole change, not just `implementation_from_dafny.rs`. So you should delete all the "smithy generated" Rust code related to that phantom operation for now too (although might be worth stashing for later). We could actually add that operation to the Smithy model, but that would break it for Java and .NET unless we also added the needed auxiliary files.

Kind of makes me think of Gen AI hallucinating an API operation that SHOULD exist but doesn't actually. :)

robin-aws (approved):

PR #389, opened 2024-05-28
ci: Disable Rust in nightly build

robin-aws:
*Description of changes:*

This workflow is pinned to a specific commit from a feature branch that's "ahead" of master, so there's no point in testing against nightly prereleases for now.


atomb (approved):
Seems reasonable. Presumably this won't be commented out for too long.

robin-aws:
>Seems reasonable. Presumably this won't be commented out for too long.

Only until we get the rest of the Dafny `feat-rust` branch merged.

PR #390, opened 2024-05-29
feat: Optionally generate (mostly) complete projects

robin-aws:
*Description of changes:*

Adds a new `--generate` option that accepts a "generation aspect" argument, and can be specified multiple times. The available aspects are:

* `project-files` - files like `build.gradle.kts` or `Service.csproj`
* `client-constructors` - externally defined [wrapped] client constructors, and similar top-level Dafny module boilerplate (open to a better name for this one)
* `impl-stubs` - stubbed out Dafny concrete operations modules and tests.

These options make it much quicker to create new libraries, especially new TestModels. The last option is intended to be used once to get started on files that do have to be manually edited, but the other two can be left on more permanently.

I set up a templated file mechanism for all these kinds of assets, since they are mostly static content with just a few pieces that depend on the service name/sdkID and other parameters. It uses [the shared `AbstractCodeWriter` templating mechanism](https://smithy.io/2.0/guides/building-codegen/generating-code.html) from the core Smithy implementation.

Also used this to generate the missing files to make `Timestamp` work for .NET (it doesn't work yet for Java because of some apparent confusion over `Date` vs. `Instant`)


TestModels/SharedMakefile.mk
```diff
@@ -55,12 +55,16 @@ _polymorph_dafny: _polymorph_wrapped
 
 _polymorph_java: OUTPUT_JAVA=--output-java $(LIBRARY_ROOT)/runtimes/java/src/main/smithy-generated
 _polymorph_java: OUTPUT_JAVA_TEST=--output-java-test $(LIBRARY_ROOT)/runtimes/java/src/test/smithy-generated
+_polymorph_java: INPUT_DAFNY=\
+		--include-dafny $(PROJECT_ROOT)/$(STD_LIBRARY)/src/Index.dfy
```
seebees, line 59: Why in both shared files?

.github/workflows/test_models_net_tests.yml
```diff
@@ -99,10 +99,6 @@ jobs:
         # https://github.com/actions/setup-dotnet/blob/main/.github/csc.json
         run: echo "::remove-matcher owner=csc::"
 
-      - name: Download Dependencies
-        working-directory: ./${{ matrix.library }}
-        run: make setup_net
-
```
seebees, line 105: Just wondering why move this? Not a big deal, just curious.

TestModels/SimpleTypes/SimpleTimestamp/test/SimpleTimestampImplTest.dfy
```diff
@@ -0,0 +1,27 @@
...
+include "../src/Index.dfy"
+
+module SimpleTimestampImplTest {
+
+    import SimpleTimestamp
+    import opened SimpleTypesTimestampTypes
+    import opened Wrappers
+
+    method {:test} TestClient(){
+        var client :- expect SimpleTimestamp.SimpleTimestamp();
```
seebees, line 12: Progress! I don't want to block, but we should at least talk about the additional tests we generally do here. What is the mechanism we can have to make sure that these tests get added?

TestModels/aws-sdks/lakeformation/runtimes/net/Extern/LakeFormationClient.cs
seebees, line 1: So this is now generated by `client-constructors` and that is why we can delete it?

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
```diff
@@ -275,6 +278,17 @@ private static Options getCliOptions() {
...
           )
           .build()
+      )
+      .addOption(
+        Option
+          .builder()
+          .longOpt("generate")
+          .desc(
+            "<optional> optional aspects to generate. Available aspects:\n" +
+            CodegenEngine.GenerationAspect.helpText()
```
seebees, line 288: Not for today, but we have a *lot* of options. We should start to plan some way to describe what they mean.

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
```diff
@@ -397,6 +410,14 @@ static Optional<CliArguments> parse(String[] args) throws ParseException {
         "update-patch-files"
       );
 
+      final String[] generationAspectOptions = Optional
+        .ofNullable(commandLine.getOptionValues("generate"))
+        .orElse(new String[0]);
+      final Set<CodegenEngine.GenerationAspect> generationAspects = Arrays
+        .stream(generationAspectOptions)
+        .map(CodegenEngine.GenerationAspect::fromOption)
+        .collect(Collectors.toSet());
```
seebees, line 420: is `--generate project-files client-constructors` a valid calling form? (Small nit, this may be the first duplicate we have in the CLI)

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
```diff
@@ -510,7 +510,8 @@ public TokenTree generateServiceTraitDefinition(ServiceShape serviceShape) {
               generateBodilessOperationMethodThatEnsuresCallEvents(
                 serviceShape,
                 operation,
-                ImplementationType.CODEGEN
+                ImplementationType.CODEGEN,
+                true
```
seebees: A naked bool at the end? Maybe. But it makes me feel icky :)

seebees: Could we incorporate this into `ImplementationType`?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyNameResolver.java
```diff
@@ -40,7 +40,9 @@ public record DafnyNameResolver(
       //            Map.entry(ShapeType.SHORT, "int16"),
       Map.entry(ShapeType.INTEGER, "int32"),
       Map.entry(ShapeType.LONG, "int64"),
-      Map.entry(ShapeType.DOUBLE, "seq<uint8>")
+      Map.entry(ShapeType.DOUBLE, "seq<uint8>"),
+      // TODO create/use better timestamp type in Dafny libraries
```
seebees, line 44: Changing this will be work, do we want to have a Datatype is basically just a string to start?

codegen/smithy-dafny-codegen/src/main/resources/templates/runtimes/net/$forLocalService:L$sdkID:L.csproj
```diff
@@ -9,24 +10,20 @@
...
-    <PackageReference Include="AWSSDK.Core" Version="3.7.100" />
-    <PackageReference Include="AWSSDK.%SDK_ID%" Version="3.7.100" />
+    <PackageReference Include="DafnyRuntime" Version="$dafnyVersion:L"/>
     <!--
       System.Collections.Immutable can be removed once dafny.msbuild is updated with
       https://github.com/dafny-lang/dafny.msbuild/pull/10 and versioned
     -->
     <PackageReference Include="System.Collections.Immutable" Version="1.7.0" />
-    <!-- Work around for dafny-lang/dafny/issues/1951; remove once resolved -->
-    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
```
seebees, line 21: Can we remove this everywhere!?!?!?!

codegen/smithy-dafny-codegen/src/main/resources/templates/src/Wrapped$service:LImpl.dfy
```diff
@@ -0,0 +1,10 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+include "../Model/$service:LTypesWrapped.dfy"
+
+module {:extern "$namespace:L.internaldafny.wrapped"} Wrapped$service:LService refines WrappedAbstract$service:LService {
```
seebees: Where are we with `outer module`? Can we use that?

seebees: add `{:options "--function-syntax:4"}`

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
```diff
@@ -313,21 +411,75 @@ private void javaAwsSdkV2(final Path outputDir) {
...
+    parameters.put("gradleDescription", gradleDescription);
+
+    if (awsSdkStyle) {
+      if (generationAspects.contains(GenerationAspect.PROJECT_FILES)) {
+        writeTemplatedFile(
+          "runtimes/java/$forSDK:Lbuild.gradle.kts",
+          parameters
+        );
+      }
+      // TODO generate sdk constructor
```
seebees, line 442: you mean the `extern` that does the native "get me an SDK"?
TBH, you could just spit it out with some throws and such so that customers know how to write it...

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
```diff
@@ -313,21 +411,75 @@ private void javaAwsSdkV2(final Path outputDir) {
...
+    } else {
+      final String serviceConfig = serviceShape
+        .expectTrait(LocalServiceTrait.class)
+        .getConfigId()
+        .getName();
+      parameters.put("serviceConfig", serviceConfig);
+
+      if (generationAspects.contains(GenerationAspect.PROJECT_FILES)) {
+        writeTemplatedFile(
+          "runtimes/java/$forLocalService:Lbuild.gradle.kts",
```
seebees, line 452: I like the `runtimes` construction, but should we be more flexible here? Is this path another one we can pass? I also suspect that we will need more fine grain. JS will likely have 2 runtimes directories even though they are the same language.

seebees (commented):
I like it. The fact that I can use them once and then not use them again makes it easy to get started.

Long term I'm still worried about the `diff` but I think that is manageable.

.github/workflows/test_models_net_tests.yml
robin-aws: If the project files are generated and not checked in, then (at least for .net) you have to generate first before you `dotnet restore`

TestModels/SimpleTypes/SimpleTimestamp/test/SimpleTimestampImplTest.dfy
robin-aws: Very good point, probably plain old testing coverage metrics on the generated code in each language?

I've also got a change in the works (split off from [this PR](https://github.com/smithy-lang/smithy-dafny/pull/257/files#diff-a933719600e0ce4e5ada8a6a7a64dfb38e005f6145c00147d9a02a16c69409d7R20-R24)) to make every language test every test model by default, and force them to loudly declare warnings for test models they aren't yet.

TestModels/aws-sdks/lakeformation/runtimes/net/Extern/LakeFormationClient.cs
robin-aws: Yes, I also intended to delete the project files but I'm in a weird state where my clone says they are already gone, but they aren't deleted in this PR...

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
robin-aws: YUP. But with #395 moving us closer to using `smithy-build.json` instead we may want to put the documentation effort there later instead.

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
robin-aws: Yes I think it's the first, I'll look at what the apache CLI library supports out of the box.

robin-aws: Looks like you can set `valueSeparator` to `,` to support `--generate client-constructors,project-files`, that sound good?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
robin-aws: Ew that is icky, shame on me for sticking that in there "temporarily" and then losing track. :P 

Another value of `ImplementationType` seems about right...

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyNameResolver.java
robin-aws: Ah I see why you thought this decision was in scope here. This is only a bug fix, we already have `string` [here](https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyNameResolver.java#L109-L110) as well, but you need it here too if you use the prelude shape directly.

codegen/smithy-dafny-codegen/src/main/resources/templates/runtimes/net/$forLocalService:L$sdkID:L.csproj
robin-aws: YES, I figured we can at least start by not having it in the template!

codegen/smithy-dafny-codegen/src/main/resources/templates/src/Wrapped$service:LImpl.dfy
robin-aws: On both points I preferred to make the templates match the current patterns as much as possible, so if we start replacing checked in files with generated ones the diffs are less scary. Consolidate, then improve :)

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
robin-aws: Yes, it just seemed less valuable to tackle right now since the KMS and DDB ones both add quite a bit of extra things, and those are the only services we test on Java now.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
robin-aws: Interesting point. For now I like the simplicity of template paths mapping directly to output paths, and it's actually quite possible we'll need different assets for each JS runtime anyway. So I think I'll say two-way door yadda yadda. :)

codegen/smithy-dafny-codegen/src/main/resources/templates/src/Wrapped$service:LImpl.dfy
seebees: I agree with outer module, with function syntax I'm more confused.

TestModels/SharedMakefile.mk
robin-aws: Because the TestModels one overrides things to always build wrapped things too, so it repeats a bunch of config :P

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
robin-aws: Huzzah, ImplementationType.DEVELOPER was sitting right there unused and exactly what I needed, see last commit!

codegen/smithy-dafny-codegen/src/main/resources/templates/src/Wrapped$service:LImpl.dfy
robin-aws: As in existing copies of this file (https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/TestModels/Errors/src/WrappedSimpleErrorsImpl.dfy) aren't doing that already, and I was trying to template the current pattern and not change it at the same time. 

But I certainly didn't check every one for consistency, and --function-syntax is small and harmless, so I'll change it. :)

codegen/smithy-dafny-codegen/src/main/resources/templates/test/$sdkID:LImplTest.dfy
```diff
@@ -0,0 +1,19 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+include "../src/Index.dfy"
+
+module $sdkID:LImplTest {
```
seebees:
```suggestion
module {:options "--function-syntax:4"} $sdkID:LImplTest {
```

seebees (approved):
LGTM

PR #393, opened 2024-05-29
Python: localService native/Dafny conversions

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
@@ -54,7 +54,9 @@ public class DafnyToLocalServiceShapeVisitor extends ShapeVisitor.Default<String
    * @param writer The PythonWriter being used
    */
   public DafnyToLocalServiceShapeVisitor(
-      GenerationContext context, String dataSource, PythonWriter writer) {
+        final GenerationContext context,
+        final String dataSource,
+        final PythonWriter writer) {
     this.context = context;
     this.dataSource = dataSource;
     this.writer = writer;
@@ -62,7 +64,10 @@ public class DafnyToLocalServiceShapeVisitor extends ShapeVisitor.Default<String
   }
 
   public DafnyToLocalServiceShapeVisitor(
-      GenerationContext context, String dataSource, PythonWriter writer, String filename) {
+        final GenerationContext context,
+        final String dataSource,
+        final PythonWriter writer,
+        final String filename) {
     this.context = context;
     this.dataSource = dataSource;
     this.writer = writer;
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
@@ -1,0 +1,319 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.amazon.polymorph.smithypython.localservice.shapevisitor;
+
+import software.amazon.polymorph.smithypython.common.nameresolver.SmithyNameResolver;
+import software.amazon.polymorph.smithypython.common.nameresolver.Utils;
+import software.amazon.polymorph.smithypython.common.shapevisitor.ShapeVisitorResolver;
+import software.amazon.polymorph.smithypython.localservice.shapevisitor.conversionwriter.DafnyToLocalServiceConversionFunctionWriter;
+import software.amazon.polymorph.smithypython.localservice.shapevisitor.conversionwriter.LocalServiceToDafnyConversionFunctionWriter;
+import software.amazon.polymorph.traits.DafnyUtf8BytesTrait;
+import software.amazon.smithy.codegen.core.CodegenException;
+import software.amazon.smithy.model.shapes.BigDecimalShape;
+import software.amazon.smithy.model.shapes.BigIntegerShape;
+import software.amazon.smithy.model.shapes.BlobShape;
+import software.amazon.smithy.model.shapes.BooleanShape;
+import software.amazon.smithy.model.shapes.ByteShape;
+import software.amazon.smithy.model.shapes.DoubleShape;
+import software.amazon.smithy.model.shapes.EnumShape;
+import software.amazon.smithy.model.shapes.FloatShape;
+import software.amazon.smithy.model.shapes.IntegerShape;
+import software.amazon.smithy.model.shapes.ListShape;
+import software.amazon.smithy.model.shapes.LongShape;
+import software.amazon.smithy.model.shapes.MapShape;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.shapes.ShapeVisitor;
+import software.amazon.smithy.model.shapes.ShortShape;
+import software.amazon.smithy.model.shapes.StringShape;
+import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.model.shapes.TimestampShape;
+import software.amazon.smithy.model.shapes.UnionShape;
+import software.amazon.smithy.model.traits.EnumTrait;
+import software.amazon.smithy.python.codegen.GenerationContext;
+import software.amazon.smithy.python.codegen.PythonWriter;
+
+/**
+ * ShapeVisitor that should be dispatched from a shape to generate code that maps a Dafny shape's
+ * internal attributes to the corresponding Smithy shape's internal attributes.
+ *
+ * <p>This generates code in a `dafny_to_smithy.py` file. The generated code consists of methods
+ * that convert from a Dafny-modelled shape to a Smithy-modelled shape. Code that requires these
+ * conversions will call out to this file.
+ */
+public class DafnyToLocalServiceShapeVisitor extends ShapeVisitor.Default<String> {
+  private final GenerationContext context;
+  private final String dataSource;
+  private final PythonWriter writer;
+  private final String filename;
+
+  /**
+   * @param context The generation context.
+   * @param dataSource The in-code location of the data to provide an output of ({@code output.foo},
+   *     {@code entry}, etc.)
+   * @param writer The PythonWriter being used
+   */
+  public DafnyToLocalServiceShapeVisitor(
+      GenerationContext context, String dataSource, PythonWriter writer) {
+    this.context = context;
+    this.dataSource = dataSource;
+    this.writer = writer;
+    this.filename = "";
+  }
+
+  public DafnyToLocalServiceShapeVisitor(
+      GenerationContext context, String dataSource, PythonWriter writer, String filename) {
+    this.context = context;
+    this.dataSource = dataSource;
+    this.writer = writer;
+    this.filename = filename;
+  }
+
+  @Override
+  protected String getDefault(Shape shape) {
+    String protocolName = context.protocolGenerator().getName();
+    throw new CodegenException(
+        String.format(
+            "Unsupported conversion of %s to %s using the %s protocol",
+            shape, shape.getType(), protocolName));
+  }
+
+  @Override
+  public String blobShape(BlobShape shape) {
+    return "bytes(%1$s)".formatted(dataSource);
+  }
+
+  @Override
+  public String structureShape(StructureShape structureShape) {
+    // ONLY write converters if the shape under generation is in the current namespace (or Unit
+    // shape)
+    if (structureShape.getId().getNamespace().equals(context.settings().getService().getNamespace())
+        || Utils.isUnitShape(structureShape.getId())) {
+      LocalServiceToDafnyConversionFunctionWriter.writeConverterForShapeAndMembers(
+          structureShape, context, writer);
+      DafnyToLocalServiceConversionFunctionWriter.writeConverterForShapeAndMembers(
+          structureShape, context, writer);
+    }
+
+    // Import the converter from where the ShapeVisitor was called
+    String pythonModuleSmithygeneratedPath =
+        SmithyNameResolver.getPythonModuleSmithygeneratedPathForSmithyNamespace(
+            structureShape.getId().getNamespace(), context);
+    writer.addStdlibImport(pythonModuleSmithygeneratedPath + ".dafny_to_smithy");
+
+    // Return a reference to the generated conversion method
+    // ex. for shape example.namespace.ExampleShape
+    // returns
+    // `example_namespace.smithygenerated.dafny_to_smithy.DafnyToSmithy_example_namespace_ExampleShape(input)`
+    return "%1$s.dafny_to_smithy.%2$s(%3$s)"
+        .formatted(
+            pythonModuleSmithygeneratedPath,
+            SmithyNameResolver.getDafnyToSmithyFunctionNameForShape(structureShape, context),
+            Utils.isUnitShape(structureShape.getId()) ? "" : dataSource);
+  }
+
+  @Override
+  public String listShape(ListShape shape) {
+    StringBuilder builder = new StringBuilder();
+
+    // Open list:
+    // `[`
+    builder.append("[");
+    MemberShape memberShape = shape.getMember();
+    final Shape targetShape = context.model().expectShape(memberShape.getTarget());
+
+    // Add converted list elements into the list:
+    // `[list_element for list_element in `DafnyToSmithy(targetShape)``
+    builder.append(
+        "%1$s"
+            .formatted(
+                targetShape.accept(
+                    ShapeVisitorResolver.getToNativeShapeVisitorForShape(
+                        targetShape, context, "list_element", writer))));
+
+    // Close structure:
+    // `[list_element for list_element in `DafnyToSmithy(targetShape)`]`
+    return builder.append(" for list_element in %1$s]".formatted(dataSource)).toString();
+  }
+
+  @Override
+  public String mapShape(MapShape shape) {
+    StringBuilder builder = new StringBuilder();
+
+    // Open map:
+    // `{`
+    builder.append("{");
+    MemberShape keyMemberShape = shape.getKey();
+    final Shape keyTargetShape = context.model().expectShape(keyMemberShape.getTarget());
+    MemberShape valueMemberShape = shape.getValue();
+    final Shape valueTargetShape = context.model().expectShape(valueMemberShape.getTarget());
+
+    // Write converted map keys into the map:
+    // `{`DafnyToSmithy(key)`:`
+    builder.append(
+        "%1$s: "
+            .formatted(
+                keyTargetShape.accept(
+                    ShapeVisitorResolver.getToNativeShapeVisitorForShape(
+                        keyTargetShape, context, "key", writer))));
+
+    // Write converted map values into the map:
+    // `{`DafnyToSmithy(key)`: `DafnyToSmithy(value)``
+    builder.append(
+        "%1$s"
+            .formatted(
+                valueTargetShape.accept(
+                    ShapeVisitorResolver.getToNativeShapeVisitorForShape(
+                        valueTargetShape, context, "value", writer))));
+
+    // Complete map comprehension and close map
+    // `{`DafnyToSmithy(key)`: `DafnyToSmithy(value)`` for (key, value) in `dataSource`.items }`
+    // No () on items call; `dataSource` is a Dafny map, where `items` is a @property and not a
+    // method.
+    return builder.append(" for (key, value) in %1$s.items }".formatted(dataSource)).toString();
+  }
+
+  @Override
+  public String booleanShape(BooleanShape shape) {
+    return dataSource;
+  }
+
+  @Override
+  public String stringShape(StringShape shape) {
+    // If shape has @DafnyUtf8BytesTrait, use bytes converter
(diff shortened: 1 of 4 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
```diff
@@ -0,0 +1,296 @@
...
+
+/**
+ * ShapeVisitor that should be dispatched from a shape to generate code that maps a Dafny shape's
+ * internal attributes to the corresponding Smithy shape's internal attributes.
+ *
+ * <p>This generates code in a `dafny_to_smithy.py` file. The generated code consists of methods
+ * that convert from a Dafny-modelled shape to a Smithy-modelled shape. Code that requires these
+ * conversions will call out to this file.
+ */
+public class DafnyToLocalServiceShapeVisitor extends ShapeVisitor.Default<String> {
```
ShubhamChaturvedi7, line 44: Nit: Why not `DafnyToSmithyShapeVisitor`?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
```diff
@@ -0,0 +1,296 @@
...
+  private final String filename;
+
+  /**
+   * @param context The generation context.
+   * @param dataSource The in-code location of the data to provide an output of ({@code output.foo},
+   *     {@code entry}, etc.)
+   * @param writer The PythonWriter being used
+   */
+  public DafnyToLocalServiceShapeVisitor(
+      GenerationContext context, String dataSource, PythonWriter writer) {
```
ShubhamChaturvedi7: Nit: Opinion varies but I prefer to finalize all the input params, local variables, etc. Everything starts with `final`, and only becomes _reference mutable_ if required.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
```diff
@@ -0,0 +1,296 @@
...
+    return "bytes(%1$s)".formatted(dataSource);
+  }
+
+  @Override
+  public String structureShape(StructureShape structureShape) {
+    // ONLY write converters if the shape under generation is in the current namespace (or Unit
+    // shape)
+    if (structureShape.getId().getNamespace().equals(context.settings().getService().getNamespace())
+        || SmithyNameResolver.isUnitShape(structureShape.getId())) {
+      LocalServiceToDafnyConversionFunctionWriter.writeConverterForShapeAndMembers(
```
ShubhamChaturvedi7, line 98: Why are we doing `LocalServiceToDafny*` Conversion here? This visitor is supposed to be for `DafnyToLocalService` right?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
```diff
@@ -0,0 +1,296 @@
...
+  }
+
+  @Override
+  public String integerShape(IntegerShape shape) {
+    return dataSource;
+  }
+
+  @Override
+  public String longShape(LongShape shape) {
+    return dataSource;
```
ShubhamChaturvedi7, line 221: I love how you don't have to handle anything here (and simultaneously crying in Go codegen)!

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
```diff
@@ -0,0 +1,296 @@
...
+        .formatted(
+            pythonModuleSmithygeneratedPath,
+            SmithyNameResolver.getDafnyToSmithyFunctionNameForShape(shape, context),
+            dataSource);
+  }
+
+  @Override
+  public String timestampShape(TimestampShape shape) {
+    // TODO-Python: This lets code generate, but will fail when code uses it
+    return "TypeError(\"TimestampShape not supported\")";
```
ShubhamChaturvedi7, line 275: Don't you need this for kms / ddb models?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/conversionwriter/DafnyToLocalServiceConversionFunctionWriter.java
```diff
@@ -0,0 +1,414 @@
...
+   * @param shape
+   * @param context
+   * @param writer
+   */
+  public static void writeConverterForShapeAndMembers(Shape shape, GenerationContext context,
+      PythonWriter writer) {
+    singleton.baseWriteConverterForShapeAndMembers(shape, context, writer);
+  }
+
+  protected void writeStructureShapeConverter(StructureShape structureShape) {
```
ShubhamChaturvedi7, line 57: `@Override` would definitely help here.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/conversionwriter/DafnyToLocalServiceConversionFunctionWriter.java
```diff
@@ -0,0 +1,414 @@
...
+          "",
+          SmithyNameResolver.getDafnyToSmithyFunctionNameForShape(unionShape, context),
+          dataSourceInsideConversionFunction,
+          () -> {
+            conversionWriter.writeComment("Convert %1$s".formatted(
+                unionShape.getId().getName()
+            ));
+
+            // First union value opens a new `if` block; others do not need to and write `elif`
+            boolean shouldOpenNewIfBlock = true;
```
ShubhamChaturvedi7, line 306: Why not use `if` for every single case? It is `Union` so only one if will match.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/conversionwriter/LocalServiceToDafnyConversionFunctionWriter.java
```diff
@@ -0,0 +1,453 @@
...
+    // DafnyStructureName(DafnyStructureMember=smithy_structure_member(...), ...)
+    // The nature of the `smithy_structure_member` conversion depends on the properties of the
+    // shape,
+    //   as described below
+    conversionWriter.writeInline("$L=", memberName);
+
+    if (context.model().expectShape(memberShape.getTarget()).hasTrait(ReferenceTrait.class)) {
+      if (memberShape.isOptional()) {
+        conversionWriter.write(
+            "((Option_Some($1L)) if (($2L is not None) and ($1L is not None)) else (Option_None())),",
```
ShubhamChaturvedi7: Nit: Might help to use first param (`$1L`) for the `dataSource` since that's the first if check and short circuits the second one.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/conversionwriter/LocalServiceToDafnyConversionFunctionWriter.java
```diff
@@ -0,0 +1,453 @@
...
+                    targetShape,
+                    context,
+                    dataSourceInsideConversionFunction + "." + CaseUtils.toSnakeCase(memberName),
+                    conversionWriter,
+                    "smithy_to_dafny")));
+      }
+
+    }
+
+    // If this is a localService config shape, defer conversion to the config ShapeVisitor TODO
```
ShubhamChaturvedi7: TODO? Why?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/conversionwriter/LocalServiceToDafnyConversionFunctionWriter.java
```diff
@@ -0,0 +1,453 @@
...
+      conversionWriter.write(
+          """
+          import $1L
+          client = $1L.default__.$2L(boto_client = $3L)
+          client.value.impl = $3L
+          return client.value
+          """,
+          DafnyNameResolver.getDafnyPythonIndexModuleNameForShape(serviceShape, context),
+          AwsSdkNameResolver.clientNameForService(serviceShape),
+          dataSourceInsideConversionFunction);
```
ShubhamChaturvedi7, line 265: Since this class is for LocalService, it will always have the `LocalServiceTrait` right?

ShubhamChaturvedi7 (commented):
Mostly style/nit/refactoring comments.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
lucasmcdonald3: I think the native conversions here are pretty specific to LocalServices.
I see Smithy as a general IDL, and the two "implementations" we have are LocalService and AWS SDK.
Both LocalService and AWS SDK are Smithy.
But this ShapeVisitor only does conversions for LocalServices.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
lucasmcdonald3: This is required for WrappedLocalService.
If some shape is only present in functions' inputs, Smithy's shape discovery process will think it is only needed for function inputs, i.e. Smithy-To-Dafny.
(Vice versa for outputs, and only generating Dafny-To-Smithy.)

This would actually be correct if we didn't have WrappedLocalService.
With WrappedLocalService, all functions need all shapes converting both ways.
Take our input-only example; a WrappedLocalService setup will need to convert its Dafny inputs to Smithy inputs (Dafny-to-Smithy). Then the LocalService does the Smithy-To-Dafny.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
lucasmcdonald3: A few things:
1. This is for localService, AWS SDKs are a different ShapeVisitor but have the same assignment
2. This literally assigns a TypeError to the field

```
someTimestampAttribute=TypeError("TimestampShape not supported")
```
so that the code generates, but we can't use that field.
3. Nothing in the ESDK Python MPL uses timestamps. I need to implement this for DBESDK though. If I had J's operation filter, I would use that here.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/conversionwriter/DafnyToLocalServiceConversionFunctionWriter.java
lucasmcdonald3: That's a good point, take a look at this

Here's as-is:

```
def aws_cryptography_materialproviders_Materials(dafny_input):
    # Convert Materials
    if isinstance(dafny_input, Materials_Encryption):
        Materials_union_value = aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.models.MaterialsEncryption(
            aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.dafny_to_smithy.aws_cryptography_materialproviders_EncryptionMaterials(
                dafny_input.Encryption
            )
        )
    elif isinstance(dafny_input, Materials_Decryption):
        Materials_union_value = aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.models.MaterialsDecryption(
            aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.dafny_to_smithy.aws_cryptography_materialproviders_DecryptionMaterials(
                dafny_input.Decryption
            )
        )
    else:
        raise ValueError("No recognized union value in union type: " + str(dafny_input))

    return Materials_union_value
```

Here's your proposal

```
def aws_cryptography_materialproviders_Materials(dafny_input):
    # Convert Materials
    if isinstance(dafny_input, Materials_Encryption):
        return  aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.models.MaterialsEncryption(
            aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.dafny_to_smithy.aws_cryptography_materialproviders_EncryptionMaterials(
                dafny_input.Encryption
            )
        )
    if isinstance(dafny_input, Materials_Decryption):
        return  aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.models.MaterialsDecryption(
            aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.dafny_to_smithy.aws_cryptography_materialproviders_DecryptionMaterials(
                dafny_input.Decryption
            )
        )
    raise ValueError("No recognized union value in union type: " + str(dafny_input))
```

Alternatively, this is also fine

```
def aws_cryptography_materialproviders_Materials(dafny_input):
    # Convert Materials
    if isinstance(dafny_input, Materials_Encryption):
        return aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.models.MaterialsEncryption(
            aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.dafny_to_smithy.aws_cryptography_materialproviders_EncryptionMaterials(
                dafny_input.Encryption
            )
        )
    elif isinstance(dafny_input, Materials_Decryption):
        return aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.models.MaterialsDecryption(
            aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.dafny_to_smithy.aws_cryptography_materialproviders_DecryptionMaterials(
                dafny_input.Decryption
            )
        )
    else:
        raise ValueError("No recognized union value in union type: " + str(dafny_input))
```

I think anything is fine here, but I'm generally biasing against making significant unnecessary changes to avoid scope creep. If you feel strongly for the latter let me know and I can change it

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/conversionwriter/LocalServiceToDafnyConversionFunctionWriter.java
lucasmcdonald3: ... I think this todo was me saying I can remove this. I had a whole sidequest around making a third set of ShapeVisitors specifically for LocalService Config shapes, but I was able to remove that eventually. I removed it and I don't see any issues...

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/conversionwriter/LocalServiceToDafnyConversionFunctionWriter.java
lucasmcdonald3: Nope -- LocalServices can visit AWS SDK service shapes. (This is part of my proposed extension to the AWS SDK TestModels; an "AWS SDK from LocalService" TestModel that does operations, errors, wrapping/unwrapping, etc. I had to discover lots of this during MPL work.)
I'll make this more clear in code

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
lucasmcdonald3: Did this in a few cases

ShubhamChaturvedi7: We should do this for local variables as well.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
ShubhamChaturvedi7: In that case we need a better naming, or at least a comment explaining this.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/conversionwriter/DafnyToLocalServiceConversionFunctionWriter.java
ShubhamChaturvedi7: I think the last one is the perfect one. But I was talking about the extra step / flag that you had to use to generate the `elif` part in the codegen (and not the final generated code). I'm okay with current implementation, just that you probably ended up putting extra effort into this :)

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/conversionwriter/LocalServiceToDafnyConversionFunctionWriter.java
```diff
@@ -131,7 +132,7 @@ private void writeNonReferenceStructureShapeMemberConverter(
     if (context.model().expectShape(memberShape.getTarget()).hasTrait(ReferenceTrait.class)) {
       if (memberShape.isOptional()) {
         conversionWriter.write(
-            "((Option_Some($1L)) if (($2L is not None) and ($1L is not None)) else (Option_None())),",
+            "((Option_Some($1L)) if (($1L is not None) and ($2L is not None)) else (Option_None())),",
```
ShubhamChaturvedi7: The swap should be for the params:

```
"((Option_Some($2L)) if (($1L is not None) and ($2L is not None)) else (Option_None())),",
           dataSourceInsideConversionFunction + "." + CaseUtils.toSnakeCase(memberName),
            targetShape.accept(
                ShapeVisitorResolver.getToDafnyShapeVisitorForShape(
                    targetShape,
                    context,
                    dataSourceInsideConversionFunction + "." + CaseUtils.toSnakeCase(memberName),
                    conversionWriter,
                    "smithy_to_dafny")));
```
otherwise you might end up with NPE since with the new change there's no short circuiting.

lucasmcdonald3: Now that I look at it again, I don't think what you're suggesting does anything.
This is *writing* code. The entire string is always written.
Swapping the parameters in codegen doesn't do anything, it just changes the order they are evaluated in codegen. But both will always be evaluated

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
lucasmcdonald3: Done

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/DafnyToLocalServiceShapeVisitor.java
lucasmcdonald3: Added comments

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/shapevisitor/conversionwriter/LocalServiceToDafnyConversionFunctionWriter.java
ShubhamChaturvedi7: Yup, it has nothing to do with the generated code. What I meant was that it's easier to read the codegen placeholder code - _If 1 is not None then check for 2_ vs if _2 is not none then check for 1._

lucasmcdonald3: Got it, done

ShubhamChaturvedi7 (approved):

Issue #356, opened 2024-05-09
Add missing configuration/features to Smithy build plugin


robin-aws:
The number and complexity of CLI arguments are getting out of hand. We already have an alternate interface into the code generation logic as a Smithy build plugin, which supports configuration via a `smithy-build.json`. It also supports standard Smithy features such as projections because it hooks into the full Smithy model building pipeline.

This task is to add the missing necessary configuration knobs to our `smithy-build.json` configuration handler so that we can migrate off the CLI.

(More specific design for how these options should map to `smithy-build.json` configuration pending...)

robin-aws:
Note the scope of this issue is only for CLI options that are relevant in `--aws-sdk` mode, since we only have the `dafny-client-plugin` plugin so far. Some options will only be needed for the `dafny-library-plugin` we have yet to create: https://github.com/smithy-lang/smithy-dafny/issues/153

robin-aws:
Here is some more detail on the scope of this task vs. future ones.

One key difference between the CLI and the Smithy build plugin is how they use input and output. The CLI has options like `--library-root` pointing to the source code directory, and options like `--output-java` that allow it to write code anywhere on disk. The Smithy CLI/Gradle plugin only gives Smithy build plugins parsed/resolved/validated Smithy models as input, and expects output to only be written to a build directory like `build/sources/dafny-client-plugin`.

I believe the best way to bridge this gap is with an additional shared Gradle plugin that can be executed after invoking the `smithy-gradle-plugin` task to build a model. This plugin would provide the connection between the contents of a directory that the Smithy build plugin outputs like `<project-root>/build/sources/dafny-client-plugin/runtimes/java/smithy-generated` and where that code is checked in, i.e. `<project-root>/runtimes/java/smithy-generated`. The `--patch-files-dir` and `--update-patch-files` features of the CLI would also move to this plugin.

The plugin would have three different tasks (or at least modes of a single task):

* `assert-same` - apply patch files, then assert that the two locations have the same content
* `update` - apply patch files, then copy from the `build` directory on top of the checked-in directory
* `update-patch-files` - don't apply patch file, update them to match the difference between the `build` directory and the checked-in copy

It looks like this is the only missing "feature" to make this work. For the record here is the analysis I did of the current set of CLI options:

## CLI options that need new smithy-build.json equivalents

* `--patch-files-dir/--update-patch-files`
    * Should be handled by post-processing

## CLI options that already have smithy-build.json equivalents

* `--model`
[...]

PR #395, opened 2024-05-29
chore: add polymorphDafny and polymorphDotnet tasks to Smithy Build Plugin-based SQS TestModel

kessplas:
*Issue #, if available:* https://github.com/smithy-lang/smithy-dafny/issues/356 (related)

*Description of changes:* This PR allows the Smithy Build Plugin-based SQS TestModel to be built and structured similarly to the Makefile/CLI based models used elsewhere. This enables some Smithy features which are not supported by the CLI, such as [Projections](https://smithy.io/2.0/guides/smithy-build-json.html#projections). 

Instead of running `make polymorph_dafny` and `make polymorph_dotnet`, run `./gradlew polymorphDafny` and `./gradlew polymorphDotnet` to generate types for the respective languages. A Makefile is included, as it is needed to run `make transpile_net` and `make test_net`, which work as expected when the above commands are executed. 

NOTE: These changes are not tested in CI. That may be done as part of this PR, or later, depending on what reviewers prefer. 


TestModels/aws-sdks/sqs/smithy-build.json
```diff
@@ -1,5 +1,23 @@
 {
   "version": "1.0",
+  "projections": {
```
robin-aws, line 3: Could we make this `listQueuesOnly` and actually hook it up as the source in the Gradle build via `projectionName = "listQueuesOnly"` as above? That way we're actually testing the workflow we're claiming to work now, and `TestComAmazonawsSqs.dfy` should still work AFAICT.

kessplas: Yeah, good call!

robin-aws: Thanks! Only nit is that "listQueuesOnly" is now technically "listQueuesOnlyExceptForAtLeastOneOtherOneBecauseOf#439" :) But it doesn't look like `smithy-build.json` supports comments. Perhaps we could add a comment to the README of this test model?

robin-aws (dismissed):

TestModels/aws-sdks/sqs/smithy-build.json
kessplas: good point, I'll do both. 
(JSON in general doesn't support comments, it's kind of a bummer)

robin-aws (approved):

Issue #369, opened 2024-05-16
Use `smithy-rs` to generate shapes for Rust


robin-aws:
Ideally we should be able to take a dependency on the `smithy-rs` plugins and use it to generate most of the types for the Rust interface to a polymorph library (see https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/TestModels/SimpleTypes/SimpleString/runtimes/rust/README.md)

It sounds like we can likely provide some kind of "no-op protocol" as a plugin and omit most of the code that isn't relevant.

Depends on https://github.com/smithy-lang/smithy-rs/issues/421 since the `smithy-rs` plugins aren't published to Maven Central yet. For now we can use a submodule and local maven publishing.

robin-aws:
This is more of an implementation idea than a task to complete - closing in favor of concrete deliverables such as https://github.com/smithy-lang/smithy-dafny/issues/375

PR #397, opened 2024-05-30
add stub for Rust Constructor TestModel

andrewbanchich:
*Issue #, if available:*

*Description of changes:*

initial example of stubbed test model so that PRs show more relevant diffs.

most of the code works, but there are still some errors, like ﻿`doesn't satisfy Stub: DafnyPrint or Stub: DafnyType`. i added no-op overrides for the Makefile rules which failed due to these.


TestModels/Constructor/Makefile
robin-aws, line 1: I think I would just leave the Makefile in the failing state, so you can at least see what doesn't compile yet easily. Since you're not also enabling this test model on Rust in CI, it doesn't have to pass initially. I think that's less confusing than something that "works" but does nothing. :)

robin-aws (changes requested):
I think this is a good idea, to make it clearer what actually changes in the PR that fixes the test model and enables it in CI.

Once I tackle #369 I think we can check in what `smithy-rs` outputs for a model initially instead, and therefore the second PR should just involve filling things like `client.rs` and `conversions`

robin-aws (approved):

PR #398, opened 2024-05-31
chore: rebuild SimpleDouble patch file

ajewellamz:
*Issue #, if available:*

*Description of changes:*


robin-aws (approved):
Hard to tell if this is right, but approving under the assumption it will fix the failure in https://github.com/smithy-lang/smithy-dafny/pull/386 :)

PR #429, opened 2024-06-04
chore : remove Config class from SimpleInteger

ajewellamz:
*Issue #, if available:*

*Description of changes:*


TestModels/SimpleTypes/SimpleInteger/runtimes/rust/src/types.rs
```diff
@@ -0,0 +1,58 @@
...
+pub struct SimpleIntegerConfig {
+    // XXX : ::std::option::Option<XXXtype>
+}
+
+impl SimpleIntegerConfig {
+    pub fn builder() -> SimpleIntegerConfigBuilder {
+	SimpleIntegerConfigBuilder::new()
+    }
+
+    pub fn to_internal(&self) -> simple_integer_dafny::_simple_dtypes_dinteger_dinternaldafny_dtypes::SimpleIntegerConfig {
```
robin-aws: I think it would be better to define these inside the `conversions` module instead, just as we do for operation inputs, outputs, and error shapes. That will keep `SimpleIntegerConfig` as a simple datatype that doesn't depend on implementation details (in this case the Dafny implementation, but even the other smithy code generators try hard to keep protocol details decoupled from shapes).

TestModels/SimpleTypes/SimpleInteger/runtimes/rust/src/types.rs
robin-aws, line 1: FWIW, `smithy-rs` only spits out module declarations directly in `types.rs`, and a separate file for each type (e.g. `types/_capture_pokemon_events.rs`. It will probably be easier to reuse codegen logic if we do the same thing?

robin-aws (changes requested):

TestModels/SimpleTypes/SimpleInteger/runtimes/rust/src/conversions/simple_integer_config/_simple_integer_config.rs
```diff
@@ -0,0 +1,27 @@
+// Code generated by software.amazon.smithy.rust.codegen.smithy-rs. DO NOT EDIT.
+#[allow(dead_code)]
+/*
+pub fn to_internal(&self) -> simple_integer_dafny::_simple_dtypes_dinteger_dinternaldafny_dtypes::SimpleIntegerConfig {
+  simple_integer_dafny::_simple_dtypes_dinteger_dinternaldafny_dtypes::SimpleIntegerConfig::SimpleIntegerConfig {}
+}
+pub fn from_internal(input : &simple_integer_dafny::_simple_dtypes_dinteger_dinternaldafny_dtypes::SimpleIntegerConfig) ->  Self {
+  Self {}
+}
+*/
```
robin-aws, line 10: Not blocking, but I don't think we actually want this in the output :)

robin-aws (approved):

PR #435, opened 2024-06-06
SimpleEnum for Rust

ajewellamz:
*Issue #, if available:*

*Description of changes:*


TestModels/SimpleTypes/SimpleEnum/codegen-patches/rust/dafny-4.5.0.patch
```diff
@@ -0,0 +1,2598 @@
...
+@@ -304,9 +276,9 @@ pub mod r#_simple_dtypes_dsmithyenum_dinternaldafny_dtypes {
+ 
+     #[derive(PartialEq, Clone)]
+     pub enum SimpleEnumShape {
+-        FIRST {},
+-        SECOND {},
+-        THIRD {},
++        FIRST,
++        SECOND,
++        THIRD,
```
robin-aws: You know what, let's NOT format `implementation_from_dafny.rs` for now, because that just creates a ton of patch noise

TestModels/SimpleTypes/SimpleEnum/runtimes/rust/src/operation.rs
```diff
@@ -0,0 +1,9 @@
+// Code generated by software.amazon.smithy.rust.codegen.smithy-rs. DO NOT EDIT.
+
+/// Types for the `GetEnum` operation.
+pub mod get_enum;
+
+/// Types for the `GetEnumKnownValue` operation.
```
robin-aws, line 6: This is a little off, there are three different operations.

Also the operations are named `GetEnumFirstKnownValueTest` etc., so we're missing the `_test` in general.

robin-aws: It's quite possible other test models have this wrong too.

robin-aws (changes requested):

TestModels/SimpleTypes/SimpleEnum/codegen-patches/rust/dafny-4.5.0.patch
ajewellamz: But I didn't. I just took off the `{}` because it makes it much harder to deal with, and that's whatb the compiler ought to produce.

TestModels/SimpleTypes/SimpleEnum/runtimes/rust/tests/simple_enum_test.rs
```diff
@@ -0,0 +1,51 @@
+use simple_enum::types::simple_enum_shape::SimpleEnumShape::*;
+use simple_enum::*;
+
+#[tokio::test]
+async fn test_get_enum() {
+    let result = client().get_enum().value(SECOND).send().await;
+    let output = result.unwrap();
+    let value = output.value().unwrap();
+    assert_eq!(value, SECOND);
```
andrewbanchich, line 9: not sure how paranoid we want to be, but this could pass if `value` is not `SECOND` based on the `PartialEq` impl for `SimpleEnumShape`. it would need to be bizarrely horrible, but it's possible.

do we want to use `match` to assert it's the expected variant?

ajewellamz: I think I'm willing to trust  `#[derive(PartialEq)]` in this case.

TestModels/SimpleTypes/SimpleEnum/codegen-patches/rust/dafny-4.5.0.patch
robin-aws: Closing the loop for posterity: after offline conversation we determined this isn't a case where it's easy for the Dafny code generator to produce the most idiomatic code, and it's not customer-facing, so we're not patching the output after all.

robin-aws (approved):
🚀

PR #436, opened 2024-06-07
feat: add Union Rust TestModel

### LINKED ISSUE #432: Benerate Union for Rust

andrewbanchich:
*Issue #, if available:*

*Description of changes:*

closes #432 


TestModels/Union/runtimes/rust/src/operation.rs
```diff
@@ -0,0 +1,7 @@
+// Code generated by software.amazon.smithy.rust.codegen.smithy-rs. DO NOT EDIT.
+
+/// Types for the `GetUnion` operation.
+pub mod get_union;
+
+/// Types for the `GetUnionKnownValue` operation.
+pub mod get_union_known_value;
```
robin-aws: Technically wrong since the Smithy operation is `GetKnownValueUnion`, but I really don't want to make you fix it manually. :) I think it will become obvious and trivial to fix when we move to actual code generation so I don't want to block on such typos in general.

robin-aws (dismissed):

TestModels/Union/runtimes/rust/src/types/_my_union.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub fn as_integer_value(&self) -> ::std::result::Result<&i32, &Self> {
+        if let MyUnion::IntegerValue(val) = &self {
+            ::std::result::Result::Ok(val)
+        } else {
+            ::std::result::Result::Err(self)
+        }
+    }
+    /// Returns true if this is a [`IntegerValue`](crate::types::MyUnion::IntegerValue).
+    pub fn is_integer_value(&self) -> bool {
+        self.as_integer_value().is_ok()
```
robin-aws, line 32: Out of curiousity, is this as efficient as something like:

```rust
if let MyUnion::IntegerValue(val) = &self {
    true
} else {
    false
}
```

If this is what smithy-rs generates then I'm sure it meets the right bar mind you.

andrewbanchich: played around with godbolt and it looks like rustc doesn't actually optimize the function call out here: https://rust.godbolt.org/z/vdhvdbf8z

robin-aws (approved):

PR #437, opened 2024-06-07
Allow Test Print

ajewellamz:
*Issue #, if available:*

*Description of changes:*


robin-aws (commented):
Is this because our current set of tests are already printing, or because we intend to start printing from them?

I generally prefer tests to not produce output, since it's only useful for local debugging and not CI runs.

ajewellamz:
Neither.
I agree that properly completed and merged tests should not print.
Back in my C++ days, I had my CI enforce that.

However, during development it is SUPER useful to be able to throw in a quick `eprintln!` to try to figure out what is happening.

As a side benefit, if our tests should not be printing, then this will at least allow the possibility that we notice that a test is printing.

robin-aws (approved):
Glad we're on the same page! Given this doesn't change whether printing is allowed, only if it actually has any effect, this is definitely a strict improvement. At least this way we can see whether tests are printing and say "hey they shouldn't be doing that" :)

PR #438, opened 2024-06-10
Benerate SimpleResources

ajewellamz:
Interesting bits :
conversions/get_resource_data/_get_resource_data_input.rs -- "new style" conversions using new routines in StandardLibrary

types/i_simple_resource.rs -- Is this what a resource should look like?


TestModels/Resource/codegen-patches/rust/dafny-4.5.0.patch
```diff
@@ -0,0 +1,782 @@
...
+             );
++            _out1.read()
+         }
+         pub fn _ctor(
+             this: &::dafny_runtime::Object<Self>,
+@@ -958,7 +959,6 @@ pub mod r#_SimpleResource_Compile {
+                 },
+             ));
+             return output.read();
+-            return output.read();
```
robin-aws: The double return is a known issue that will be fixed soon at least, so reasonable to fix it. :) Was it a compile error?

TestModels/Resource/codegen-patches/rust/dafny-4.5.0.patch
```diff
@@ -0,0 +1,782 @@
...
+index 4aadfc53..f20e4319 100644
+--- b/TestModels/Resource/runtimes/rust/dafny_impl/src/implementation_from_dafny.rs
++++ a/TestModels/Resource/runtimes/rust/dafny_impl/src/implementation_from_dafny.rs
+@@ -548,7 +548,7 @@ pub mod r#_simple_dresources_dinternaldafny_dtypes {
+             }
+         }
+     }
+-
++/*
+     impl ::std::default::Default for GetResourcesOutput {
```
robin-aws: Is this broken in some way? It's not obvious to me just glancing at the patched file.

Feel free to add comments when you have to do this, even links to Dafny issues

TestModels/Resource/codegen-patches/rust/dafny-4.5.0.patch
```diff
@@ -0,0 +1,782 @@
...
+             output = ::dafny_runtime::MaybePlacebo::from(_out0.read());
+             return output.read();
+@@ -845,13 +845,14 @@ pub mod r#_SimpleResource_Compile {
+                 ::std::rc::Rc<super::r#_simple_dresources_dinternaldafny_dtypes::Error>,
+             >,
+         > {
+-            let mut _out1 = ::dafny_runtime::MaybePlacebo::<::std::rc::Rc<super::r#_Wrappers_Compile::Result<::std::rc::Rc<super::r#_simple_dresources_dinternaldafny_dtypes::GetResourceDataOutput>, ::std::rc::Rc<super::r#_simple_dresources_dinternaldafny_dtypes::Error>>>>::new();
+-            _out1 = ::dafny_runtime::MaybePlacebo::from(
++            // let mut _out1 = ::dafny_runtime::MaybePlacebo::<::std::rc::Rc<super::r#_Wrappers_Compile::Result<::std::rc::Rc<super::r#_simple_dresources_dinternaldafny_dtypes::GetResourceDataOutput>, ::std::rc::Rc<super::r#_simple_dresources_dinternaldafny_dtypes::Error>>>>::new();
++            let _out1 = ::dafny_runtime::MaybePlacebo::from(
```
robin-aws: Was this broken, or did it just seem ridiculous to declare a variable with a placebo value that is then immediately overwritten? :) 

If the latter you can leave it, another case where the current approach of code generation always works but looks dumb in degenerate cases.

TestModels/Resource/codegen-patches/rust/dafny-4.5.0.patch
```diff
@@ -0,0 +1,782 @@
...
++*/
+     impl ::std::convert::AsRef<GetResourcesOutput> for &GetResourcesOutput {
+         fn as_ref(&self) -> Self {
+             self
+@@ -588,7 +588,7 @@ pub mod r#_simple_dresources_dinternaldafny_dtypes {
+             let mut output = ::dafny_runtime::MaybePlacebo::<::std::rc::Rc<super::r#_Wrappers_Compile::Result<::std::rc::Rc<super::r#_simple_dresources_dinternaldafny_dtypes::GetResourceDataOutput>, ::std::rc::Rc<super::r#_simple_dresources_dinternaldafny_dtypes::Error>>>>::new();
+             let mut _out0 = ::dafny_runtime::MaybePlacebo::<::std::rc::Rc<super::r#_Wrappers_Compile::Result<::std::rc::Rc<super::r#_simple_dresources_dinternaldafny_dtypes::GetResourceDataOutput>, ::std::rc::Rc<super::r#_simple_dresources_dinternaldafny_dtypes::Error>>>>::new();
+             _out0 = ::dafny_runtime::MaybePlacebo::from(
+-                ::dafny_runtime::md!(self).r#_GetResourceData_k(input),
++                self.r#_GetResourceData_k(input)
```
robin-aws: I assume this one didn't compile without the call to `md!` so this looks good.

robin-aws (commented):
Commented on the implementation_from_dafny.rs patches.

I think in general we should start including comments that link to Dafny issues whenever we have to correct the code generation.

TestModels/Resource/runtimes/rust/src/types/i_simple_resource.rs
```diff
@@ -0,0 +1,56 @@
...
+        &mut self,
+        input: crate::operation::get_resource_data::GetResourceDataInput,
+    ) -> Result<
+        crate::operation::get_resource_data::GetResourceDataOutput,
+        crate::operation::get_resource_data::GetResourceDataError,
+    >;
+}
+
+#[derive(::std::clone::Clone, ::std::cmp::PartialEq, ::std::fmt::Debug)]
+pub struct ISimpleResourceObject {
```
robin-aws: Can we reduce the visibility of the implementing struct somehow, since customer code should not have to reference it directly?

I'm also inclined to name it `SimpleResourceImpl` instead as "object" doesn't convey the right meaning for me.

TestModels/Resource/runtimes/rust/src/types/i_simple_resource.rs
```diff
@@ -0,0 +1,56 @@
+// Code generated by software.amazon.smithy.rust.codegen.smithy-rs. DO NOT EDIT.
+
+pub trait ISimpleResource {
```
robin-aws: The "I" prefix doesn't seem very idiomatic for Rust, I think we should drop it.

TestModels/Resource/runtimes/rust/src/conversions/get_resource_data/_get_resource_data_input.rs
```diff
@@ -0,0 +1,39 @@
...
+    value: crate::operation::get_resource_data::GetResourceDataInput,
+) -> ::std::rc::Rc<
+    ::simple_resources_dafny::r#_simple_dresources_dinternaldafny_dtypes::GetResourceDataInput,
+> {
+    ::std::rc::Rc::new(::simple_resources_dafny::r#_simple_dresources_dinternaldafny_dtypes::GetResourceDataInput::GetResourceDataInput {
+        stringValue: dafny_standard_library::conversion::ostring_to_dafny(value.string_value()),
+        booleanValue: dafny_standard_library::conversion::obool_to_dafny(value.boolean_value()),
+        integerValue: dafny_standard_library::conversion::oint_to_dafny(value.integer_value()),
+        longValue: dafny_standard_library::conversion::olong_to_dafny(value.long_value()),
+        blobValue: dafny_standard_library::conversion::oblob_to_dafny(value.blob_value())
```
robin-aws, line 13: Love the conversion helper methods btw :)

ajewellamz: Thanks.

TestModels/Resource/runtimes/rust/src/types/i_simple_resource.rs
ajewellamz: Done

TestModels/Resource/runtimes/rust/src/operation/get_resources/_get_resources_output.rs
```diff
@@ -0,0 +1,66 @@
+// Code generated by software.amazon.smithy.rust.codegen.smithy-rs. DO NOT EDIT.
+#[allow(missing_docs)] // documentation missing in model
+#[non_exhaustive]
+#[derive(::std::clone::Clone, ::std::cmp::PartialEq, ::std::fmt::Debug)]
+pub struct GetResourcesOutput {
+    pub(crate) output: crate::types::simple_resource::SimpleResourceImpl,
```
robin-aws: So my instincts tell me this should somehow only refer to the `SimpleResource` trait instead. Is there a way to produce some box of a `dyn SimpleResource` instead without unacceptable overhead? I don't know Rust quite well enough to know the exact solution but I strongly suspect it exists.

This will matter more when we get to the `Extendable` test case, where `@extendable` allows for customer implementations of resources, and therefore HAS to be something like what I'm suggesting. It should probably be backwards compatible to add `@extendable` to a resource, but if non-extendable resources are referenced by their concrete type, it wouldn't work to change it to the trait later.

ajewellamz: The trouble is converting between that and the Dafny type, what with ownership and whatnot.
I'll bang my head against it a little more.

robin-aws: Converting might have to make use of some unsafe-ly implemented routines from the Dafny runtime, so don't let the conversion difficulties convince us it's not possible.

The key point is we only want to promise that `GetResources` returns something that implements the trait, and not the specific struct type.

As a strawperson idea, could we just make it `Box<dyn SimpleResource`?

ajewellamz: Yes, that's the direction I'm going. 
Box didn't work, so I'm using `Rc<RefCell<dyn SimpleResource>>`

robin-aws: Great, that's definitely better. I'm not 100% confident there isn't a better choice that doesn't force reference counting, but I'd rather move forward with this choice if it's correct even if it isn't maximally efficient.

I cut https://github.com/smithy-lang/smithy-dafny/issues/449 as a follow up we should do to check whether some of these kinds of choices will work with typical uses.

robin-aws (approved):

PR #440, opened 2024-06-11
Benerate SimpleErrors

### LINKED ISSUE #404: Benerate Errors for Rust

ajewellamz:
*Issue #, if available:*
Resolves #404 

*Description of changes:*


robin-aws:
Pushed a commit to refactor `always_error` to use the approach to errors I proposed in #450 (it was a lot of change to do one operation so I don't want to go ahead with the other two immediately). The key places I'm trying to illustrate how it works are:

* The `DafnyError` struct in the standard library, which can wrap up an arbitrary Dafny value as a `std::error::Error` without losing information: https://github.com/smithy-lang/smithy-dafny/pull/440/files#diff-0b48ab31781d189bf728e6287b703be78a335da810978ca454e0cb263051ea09R109
* Restoring the per-operation error types, where each has an explicit variant for each modeled error (in this case only `SimpleErrorsException`: https://github.com/smithy-lang/smithy-dafny/pull/440/files#diff-9c199f24fbed164fad9a89e2b45192ec3194baeb01855a6d710491e90849f027R41
* Converting between a Dafny `Error` and a `AlwaysErrorError` without losing information: https://github.com/smithy-lang/smithy-dafny/pull/440/files#diff-56b16340a27f1927951bc4d032b25cc2d26abb5f09088753499a36e526e68237

robin-aws:
Reverted my commit to move in the direction of #450 - see https://github.com/smithy-lang/smithy-dafny/issues/450#issuecomment-2305866211

TestModels/Errors/runtimes/rust/src/types/error.rs
```diff
@@ -0,0 +1,45 @@
+// Code generated by software.amazon.smithy.rust.codegen.smithy-rs. DO NOT EDIT.
+
+#[derive(Clone, Debug)]
+pub struct Object(pub Option<::std::rc::Rc<std::cell::UnsafeCell<dyn ::std::any::Any>>>);
```
robin-aws: Could we use `dafny_runtime::Object` directly instead?

ajewellamz: Yes, it seems like we should use `Object` instead. I have no idea what I was thinking.

robin-aws (approved):

PR #441, opened 2024-06-12
Python: Integrate with CodegenEngine and CodegenCli

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
```diff
@@ -579,6 +647,22 @@ public Builder withNamespace(final String namespace) {
...
+     */
+    public Builder withDependencyModuleNames(final Map<String, String> dependencyModuleNames) {
+      this.dependencyModuleNames = dependencyModuleNames;
+      return this;
+    }
+
+    /**
+     * Sets the Python module name for any generated Python code.
+     */
+    public Builder withModuleName(final String moduleName) {
```
robin-aws: Could we just derive this from the Smithy service shape name? Do we actually need this level of control?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
```diff
@@ -579,6 +647,22 @@ public Builder withNamespace(final String namespace) {
       return this;
     }
 
+    /**
+     * Sets the directories in which to search for dependent model file(s).
+     */
+    public Builder withDependencyModuleNames(final Map<String, String> dependencyModuleNames) {
```
robin-aws: See below, if this is also only necessary because we are supporting customizing the module name in general, I'm hoping this isn't necessary either.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
lucasmcdonald3: We cannot derive this from the Smithy service shape name.
Take the MPL-Core as an example.
The MPL-Core has two Smithy services inside of it, `aws.cryptography.materialProviders` and `aws.cryptography.keyStore`.
Both of these are in the `aws_cryptographic_materialproviders` module.
For `aws.cryptography.materialProviders`, we *might* be able to change something to get everything working.
But for `aws.cryptography.keyStore`, it really falls apart.

robin-aws: I got it now, forgot about the fact that a polymorph library can have more than one service in it. And now that we've chatted more about the related-but-separate issue of `--outer-module`/`--python-module-name`, I know what we should do here:

We should call this `--library-name`, and expect something like `aws.cryptography.materialProviders`. smithy-dafny will map that value to an idiomatic value for the target language's "outer namespace" option, such as `--python-module-name` or (as we discussed just now) `--java-package` or whatever we call it.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
robin-aws: Still not clear if this is needed given my suggestion above - if we have a strong smithy-dafny convention for things, does it become unnecessary?

ShubhamChaturvedi7: It can't be fully derived using convention in Go - unless you really make it a hard convention. Example Go Module Name: `github.com/aws/<aws-sdk>` vs `github.com/smithy-lang/<TestModels>` . Go also needs domain name.

lucasmcdonald3: Something like this is still necessary, as Shubham indicated.
Since there is no clean way to derive this, we should provide some configurable naming scheme like this.
I think there's a second iteration of this solution where a developer doesn't have to provide this map in every project, but that would take significantly more work.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
lucasmcdonald3: Renamed the generic variable in these files to `libraryName`. Python takes this as `module` names since Python uses modules, but Java and NET can grab this and call their particular strings whatever works for them.

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
```diff
@@ -171,6 +184,14 @@ private static Options getCliOptions() {
           .required()
           .build()
       )
+      .addOption(
+        Option
+          .builder("ln")
+          .longOpt("library-name")
+          .desc("if generating for a language that uses library names (go, python), the name of the library")
```
seebees: in this case `uses` ==> `dependency`?

lucasmcdonald3: I'm not following...
Maybe some examples will answer your question?
```
AwsCryptographyPrimitives: python-module-name --> library-name = aws_crytography_primitives
AwsCryptographicMaterialProviders: python-module-name --> library-name = aws_crytographic_materialproviders
```

seebees: What about:

`if generating for a language that uses library names (go, python), the name of the library in that language`

seebees (approved):
LGTM

PR #442, opened 2024-06-12
feat: Include Java runtime version in project.properties as well

robin-aws:
*Description of changes:*

For the benefit of the DB ESDK nightly build, since it defines `dafnyRuntimeJavaVersion` separately from `dafnyVersion`.

It also turns out I accidentally dropped handling of the `--properties-file` path in #390, so this also allows `writeTemplatedFile` to customize the output path.


ajewellamz (approved):

seebees (approved):
LGTM

PR #448, opened 2024-06-13
Python: WrappedLocalService Smithy-Python extensions

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/wrappedlocalservice/extensions/DafnyPythonWrappedLocalServiceClientCodegenPlugin.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/wrappedlocalservice/extensions/DafnyPythonWrappedLocalServiceClientCodegenPlugin.java
@@ -1,0 +1,96 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.amazon.polymorph.smithypython.wrappedlocalservice.extensions;
+
+import java.util.Map;
+import software.amazon.polymorph.smithypython.common.nameresolver.SmithyNameResolver;
+import software.amazon.polymorph.traits.WrappedLocalServiceTrait;
+import software.amazon.polymorph.traits.LocalServiceTrait;
+import software.amazon.smithy.build.PluginContext;
+import software.amazon.smithy.build.SmithyBuildPlugin;
+import software.amazon.smithy.codegen.core.directed.CodegenDirector;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.transform.ModelTransformer;
+import software.amazon.smithy.python.codegen.GenerationContext;
+import software.amazon.smithy.python.codegen.PythonSettings;
+import software.amazon.smithy.python.codegen.PythonWriter;
+import software.amazon.smithy.python.codegen.integration.PythonIntegration;
+import software.amazon.smithy.utils.SmithyUnstableApi;
+
+/**
+ * Plugin to trigger Smithy-Dafny Python code generation for a wrapped localService.
+ * This differs from the PythonClientCodegenPlugin by not calling `runner.performDefaultCodegenTransforms()`
+ * and `runner.createDedicatedInputsAndOutputs()`.
+ *
+ * This differs from the non-wrapped plugin by adding a WrappedLocalServiceTrait
+ * to the service that is being generated. This signals that this plugin requires wrapped localService
+ * generation, so that we can identify this from within code generation.
+ */
+@SmithyUnstableApi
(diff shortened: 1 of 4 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/wrappedlocalservice/extensions/DafnyPythonWrappedLocalServiceClientCodegenPlugin.java
```diff
@@ -0,0 +1,96 @@
...
+                  throw new IllegalArgumentException(
+                      "ServiceShape for LocalService test MUST have a LocalServiceTrait: "
+                          + serviceShape);
+                }
+                return serviceShape.toBuilder()
+                    .addTrait(WrappedLocalServiceTrait.builder().build())
+                    .build();
+              } else {
+                return shape;
+              }
```
texastony, line 65: Question: Won't this mark every Service Shape with a WrappedLocalServiceTrait, as compared to just the "subject".
For example, if I wanted to just Wrap the MPL, and not the Primitives, which is a model that it consumes, this logic would appear to add the trait to both.

texastony: Oh... `serviceShape` is an input... do you have to use a model transformer?
Why not just?:
```
               if (!shape.hasTrait(LocalServiceTrait.class)) {
                  throw new IllegalArgumentException(
                      "ServiceShape for LocalService test MUST have a LocalServiceTrait: "
                          + serviceShape);
                }
                return serviceShape.toBuilder()
                    .addTrait(WrappedLocalServiceTrait.builder().build())
                    .build();
```

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/wrappedlocalservice/extensions/DafnyPythonWrappedLocalServiceSymbolVisitor.java
```diff
@@ -0,0 +1,53 @@
+package software.amazon.polymorph.smithypython.wrappedlocalservice.extensions;
```
texastony, line 1: Nit: Missing Copyright header. 
Is there no CI check for this?
```suggestion
// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
// SPDX-License-Identifier: Apache-2.0

package software.amazon.polymorph.smithypython.wrappedlocalservice.extensions;
```

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/traits/WrappedLocalServiceTrait.java
texastony, line 1: Praise: This trait is great!
You should share it with the Rust and Go devs,
and we should try and get it into .NET and Java.

texastony (approved):

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/wrappedlocalservice/extensions/DafnyPythonWrappedLocalServiceClientCodegenPlugin.java
lucasmcdonald3: I haven't done much testing here, but I think I need to use the Model Transformer to change the `serviceShape` that is inside the model

texastony: That is my point... this is going to change every `serviceShape` in  the model, 
not ONLY the `serviceShape` provided.

Or am I reading this transformer incorrectly?

lucasmcdonald3: No, this only changes the `serviceShape` that was provided as this function's argument.
The `execute` method passes in the `serviceShape` under generation, so in practice this only changes the serviceShape under generation.

I can't use the code snippet you provided because the serviceShape inside the model object and the one returned by your code snippet are two different objects in memory.
I also need to overwrite the shape inside the model to have this trait.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/wrappedlocalservice/extensions/DafnyPythonWrappedLocalServiceSymbolVisitor.java
lucasmcdonald3: Added.
There may be a CI check for this, but I'm not focusing on getting the CI green right now. That will come in a separate PR

Issue #373, opened 2024-05-16
Add TestModel for recursive shapes (add to Aggregate)


robin-aws:
Create a TestModel that uses a "recursive" union or structure; e.g.

```smithy
structure RecursiveStructure {
    structureMember: RecursiveStructure
}

union RecursiveUnion {
    unionMember: RecursiveUnion
}
```
(Or, augment the Aggregate TestModel.)

This is not validated in the TestModels, but is used in the MPL/DBESDK.
This is particularly important for codegens that are built on other Smithy code generators. Those code generators push developers toward tools that do not handle this at all.

robin-aws:
Note AggregateReferences has some similar shapes, but that test model is mostly aimed at exercising the generation of Dafny specifications.

robin-aws:
Another thing Aggregate doesn't have coverage for is `@required` aggregate members.

PR #459, opened 2024-06-24
Benerate Aggregate

ajewellamz:
*Issue #, if available:*

*Description of changes:*


TestModels/Aggregate/runtimes/rust/src/operation/get_aggregate/_get_aggregate_output.rs
```diff
@@ -0,0 +1,214 @@
...
+    #[allow(missing_docs)] // documentation missing in model
+    pub simple_integer_map:
+        ::std::option::Option<::std::collections::HashMap<::std::string::String, i32>>,
+    #[allow(missing_docs)] // documentation missing in model
+    pub nested_structure: ::std::option::Option<crate::types::NestedStructure>,
+}
+impl GetAggregateOutput {
+    #[allow(missing_docs)] // documentation missing in model
+    ///
+    /// If no value was sent for this field, a default will be set. If you want to determine if no value was sent, use `.simple_string_list.is_none()`.
```
robin-aws: Interesting, I see that this is what `smithy-rs` does, but to me it seems just as dangerous as it is convenient. I think it's worth getting background on this decision from @rcoh

robin-aws: I do see it's controlled by a `flattenCollectionAccessors` config setting that seems to be off by default.

Edit: Off by default in server codegen, but on by default in client codegen.

rcoh: yeah—this is most useful for AWS models that essentially unchangeable and have incorrect modeled behavior (for example, a collection in return position modeled as optional)—it is very unlikely that you as a consumer cared about that distinction.

If you control the model, then this isn't necessary–because you should model it as required!

rcoh: RFC: https://github.com/smithy-lang/smithy-rs/blob/main/design/src/rfcs/rfc0035_collection_defaults.md

robin-aws: Thanks for the background @rcoh! So sounds like we need to be prepared to adjust for this when wrapping a Rust AWS SDK into Dafny, but for our own local service models we should just go by whether the member is `@required` or not.

I also made a note on https://github.com/smithy-lang/smithy-dafny/issues/373 that we should add coverage of required aggregate members to the test model (after getting this merged)

TestModels/Aggregate/runtimes/rust/tests/aggregate.rs
robin-aws, line 1: I see we're missing the `TestGetAggregateKnownValue` equivalent, but I think there's diminishing returns on writing that manually, and we can hopefully compile the actual Dafny tests soon enough to not bother with that.

robin-aws (approved):
Nice!

PR #460, opened 2024-06-26
feat: Support Dafny tests on test models for Rust

### LINKED ISSUE #458: Support Dafny tests for benerated test models
We've been manually editing Rust source tests, which is also valuable testing, but it's also very easy to test for the wrong symbols.

Primary blocker is figuring out the target for the Dafny tests - they are translated into a single main method instead of individual `[#test]`s. What's the most idiomatic way to provide a main method that should only be used for testing a library and not called by consumers?

robin-aws:
*Issue #, if available:*
Resolves #458

*Description of changes:*

Supports compiling the common Dafny tests in a test model for Rust. This involved having to change some properties of how we lay out the Rust crates:

* ~Removed the `dafny_impl` sub-crate and moved `implementation_from_dafny.rs` into the main `src` directory - hence replacing `::simple_boolean_dafny` with `crate::implementation_from_dafny` everywhere. I originally had this separated to better divide Dafny-generated code from Smithy-generated code, but it made implementing externs hard/impossible, and Dafny tests make use of "wrapped services" which are essentially testing-only extern shims.~
  * This turned out to be unnecessary, because it is reasonable to patch in additional `use ::simple_boolean_dafny::*;` and 
`use simple_boolean::*;` imports into the compiled tests. This is equivalent to the `pub use dafny_standard_library::implementation_from_dafny::*;` import the Makefile is adding, and will be replaced by a Dafny feature named something like `--rust-module-name` as some other other supported languages already have.
* Added ~`tests/tests_from_dafny/_wrapped.rs`~ `wrapped::client::Client` with the implementation of the "wrapped service", which is implementing the Dafny-client interface using the Rust idiomatic client.
  * Since this is only used for testing, but implements a Dafny extern that is only defined by Dafny test code, I guarded this with a `wrapped-client` feature which the Dafny-compiled integration test enables, as per https://github.com/rust-lang/cargo/issues/2911#issuecomment-749580481
* ~Removed `async` from the client interfaces - we'd originally kept these for better forwards-compatibility, but AFAICT it's impossible to implement the synchronous Dafny-generated trait methods with async methods. This just means that in the future we'll eventually have to provide separate async clients, but that's happened with the AWS SDKs frequently as well.~
  * Worked around this by instantiating a `current_thread` Tokio `Runtime` as per https://tokio.rs/tokio/topics/bridging instead. 
* ~Raised the visibility of a few things in the main crate to `pub` so that the `dafny_tests` integration test crate can see them - open to feedback here on whether it's reasonable or we should find a better solution.~
  * Not necessary any more, see edit to the bullet about the wrapped client.

~Not ready to merge yet because this relies on a newer Dafny feature branch commit, which I'll tackle in a separate PR first since it will be noisy to update all the patch files.~


TestModels/SimpleTypes/SimpleBoolean/runtimes/rust/tests/tests_from_dafny/_wrapped.rs
robin-aws, line 1: The error handling code in here is all over the map, please help me use some cleaner idioms :) I can't use `?` in general since `r#_Wrappers_Compile::Result` is not a Rust `Result`, but perhaps we can use `Result.map_err` somehow?

TestModels/SimpleTypes/SimpleBoolean/runtimes/rust/src/wrapped/client.rs
robin-aws, line 1: The error handling code in here is all over the map, please help me use some cleaner idioms :) I can't use `?` in general since `r#_Wrappers_Compile::Result` is not a Rust `Result`, but perhaps we can use `Result.map_err` somehow?

robin-aws: Did the best I could with this. :)

andrewbanchich (approved):

Issue #359, opened 2024-05-13
[.NET] Support @documentation trait only for plaintext content


texastony:
### Issue
Currently, the Java generated code contains documentation 
from the Smithy Model that is annotated with `@javadoc`.

.NET generated code does not.

Thus, products that use Smithy-Dafny for .NET do not get API documentation,
AND these products are harder to use in .NET.

### Fix

Since `@javadoc` only supports plaintext content,
we can decide that smithy-dafny supports the standard `@documentation` trait
but only if it does not contain any markup.
We can add a validator that produces `WARNING` events
for any `@documentation` traits that contain markup.
Then we can just replace `@javadoc` with `@documentation` in existing models
and deprecate the former.

### Original proposed fix:

Refactor the .NET generation logic to respect  `@javadoc`.

This is related to https://github.com/smithy-lang/smithy-dafny/issues/247,
but does not go as far.

robin-aws:
While it seems very weird to support a trait named `@javadoc` on other languages, I understand the motivation here that existing projects have used `@javadoc` with plaintext content that doesn't use any CommonMark markup, and therefore is fine to use for other languages.

I think instead that smithy-dafny should support `@documentation` but with a validator that asserts that the content of each instance doesn't have any markup, and replace `@javadoc` with `@documentation` in existing models.

texastony:
I am happy with that proposal!!
If #247 is refactored to your proposed scope, 
this issue can be resolved as a duplicate.

texastony:
Alternatively, #247 can be left alone and we could revise this issue to be your proposal.

robin-aws:
Yes let's do that, since it would still be good to fully support `@documentation` with markup in the future.

PR #463, opened 2024-06-27
feat: Support @documentation (for Java initially)

### LINKED ISSUE #427: Add TestModel for @documentation (Documentation)

robin-aws:
*Issue #, if available:*
Resolves #427. Progresses on #359.
 
*Description of changes:*
Adds a validator that any use of `@documentation` on local services does not use any CommonMark markup, and augments the Java code generation to recognize `@documentation` anywhere it currently recognized `@javadoc`.

Also adds a test model for validation and Java for now.


codegen/smithy-dafny-codegen/src/main/resources/META-INF/services/software.amazon.smithy.model.validation.Validator
```diff
@@ -1 +1,2 @@
-software.amazon.polymorph.traits.ReferenceTraitValidator
+software.amazon.polymorph.traits.NoReferencesInSmokeTestsValidator
```
robin-aws, line 1: Leftover fix from https://github.com/smithy-lang/smithy-dafny/pull/355#issuecomment-2105386798

TestModels/Documentation/Model/documentation.smithy
```diff
@@ -0,0 +1,92 @@
...
+  message: String
+}
+
+/// Either kind of thing.
+union SomeKindOfThing {
+  thing: Thing
+  widget: WidgetReference
+}
+
+/// Types of widgets you can get.
```
ajewellamz, line 100: How about a multi-line comment?

TestModels/Documentation/Model/documentation.smithy
```diff
@@ -0,0 +1,92 @@
...
+@enum([
+  {
+    name: "BLUE",
+    value: "BLUE"
+  },
+  {
+    name: "LARGE",
```
ajewellamz, line 107: Comments on the individual values?

TestModels/Documentation/Model/documentation.smithy
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/// Outputs for getting a thing.
+structure GetThingOutput {
+  /// The thing that you just got.
+  @required
+  thing: Thing
+}
+
+/// A different kind of thing you can get.
```
ajewellamz: Where in the spec does it say `///` is a comment? It clearly is, but I can't find the spec.

robin-aws: https://smithy.io/2.0/spec/idl.html#documentation-comment

TestModels/Documentation/Model/documentation.smithy
texastony, line 1: Issue/Question: Where is the `@documentation` trait being applied?

TestModels/Documentation/Model/documentation.smithy
texastony: What! Why?
That... ugh...
At least one of of these `///` should be replaced with the actual trait instance.
https://smithy.io/2.0/spec/documentation-traits.html#smithy-api-documentation-trait

Maybe I am crazy, but I hope that our team always uses the actual trait,
instead of `///`.

`///` is too easy to confuse with `//` or `//#`,
which are just regular comments
and duvet comments respectively.

TestModels/Documentation/Model/documentation.smithy
```diff
@@ -0,0 +1,92 @@
...
+
+structure SetWidgetNameInput {
+  @required
+  name: String
+}
+
+/// Error returned when we couldn't get the thing.
+@error("server")
+structure CouldntGetTheThingError {
+  /// Error message (this documentation is ignored!)
```
texastony: REALLY! This is ignored!
Yeah... fuck this `///` syntax.

codegen/smithy-dafny-codegen-test/bin/test/software/amazon/smithy/dafny/codegen/TestModelTest.class
texastony, line 1: Err... did you mean to check this in?
I may have fallen behind on Smithy-Dafny conventions,
but I did not know we were checking in binaries/compiled files.

texastony (commented):
We need at least one instance of the actual trait,
instead of the `///` short cut that I find... questionable...

TestModels/Documentation/Model/documentation.smithy
robin-aws: Good call, included that in the unit test but not the model. And was handling multiple paragraphs but not multiple lines in paragraphs. Updated.

TestModels/Documentation/Model/documentation.smithy
robin-aws: That's a separate mechanism from the `@document` trait (`documentation: "Blah blah"`), and we don't support that yet. It would actually be better to support the Smithy 2.0 `enum` shape instead, because you CAN apply `@documentation` directly to an enum shape's members.

I was going for parity with what Java supports, and therefore what the actual smithy-dafny projects are using.

codegen/smithy-dafny-codegen-test/bin/test/software/amazon/smithy/dafny/codegen/TestModelTest.class
robin-aws: NO, just realized locally this was showing up as modified :P

TestModels/Documentation/Model/documentation.smithy
robin-aws: It would happen with `@documentation` as well, the issue is OUR code generation is special casing `message` on errors and ignoring it.

I'll be more specific in the comment.

robin-aws:
> We need at least one instance of the actual trait,
> instead of the /// short cut that I find... questionable...

The `/// ...` syntax is pure sugar for `@documentation("...")`, and by the time Smithy parses the model and gives it to us there's no difference. I can add at least one explicit `@documentation` trait just for completeness and help explain, but there's literally no difference in the model.

We can debate separately whether we want to use that syntax in our models. Personally I'd prefer it because it improves readability.

I'll also add background to the `README.md` (which I just realized I left out).

TestModels/Documentation/Model/documentation.smithy
robin-aws: I admit I AM disappointed that the VS Code Smithy extension doesn't give `///` comments a different color or something. :P

TestModels/Documentation/Model/documentation.smithy
texastony: OH. OK. 
If this was any other shape, it work, right?
For example:
```
@javadoc("Configures Key Store's KMS Key ARN restrictions.")
union KMSConfiguration {
  @documentation("<blah>")
  // The Key Store (released on 2023-07-24) only allowed for this.
  kmsKeyArn: com.amazonaws.kms#KeyIdType,

  ///<more-blah>
  kmsMRKeyArn: com.amazonaws.kms#KeyIdType,
}
```
\<blah> and \<more-blah> would be detected, correct?

Because they are right now with JavaDoc.
Evidence:
- [Smithy Model](https://github.com/aws/aws-cryptographic-material-providers-library/blob/633e4c1639fd3fb5725779943547b751b647d6a7/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/Model/KeyStore.smithy#L108-L125)
- [Smithy Generated Java](https://github.com/aws/aws-cryptographic-material-providers-library/blob/633e4c1639fd3fb5725779943547b751b647d6a7/AwsCryptographicMaterialProviders/runtimes/java/src/main/smithy-generated/software/amazon/cryptography/keystore/model/KMSConfiguration.java#L8-L31)

robin-aws: Ah definitely. And the test model has quite a few cases of documentation on members.

codegen/smithy-dafny-codegen/src/test/java/software/amazon/polymorph/traits/ValidatorTest.java
```diff
@@ -0,0 +1,38 @@
...
+package software.amazon.polymorph.traits;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.commonmark.node.Node;
+import org.commonmark.parser.Parser;
+import org.junit.Test;
+
+public class ValidatorTest {
```
texastony, line 13: Nit: Weird name for this class, but I am not blocking on it.

texastony (approved):

codegen/smithy-dafny-codegen/src/test/java/software/amazon/polymorph/traits/ValidatorTest.java
robin-aws: Fair, I was thinking we'll likely want more validator unit tests in the future. Will improve at next opportunity.

PR #466, opened 2024-07-02
fix(Java-SDK): handle generic service exceptions

texastony:
*Issue #, if available:*

SDK Exceptions that are not modeled by a service are treated as Opaque.
In Java, this lead to an IllegalStateException, instead of a NotAuthorized Exception.
This PR refactors the generated Java to capture un-modeled SDK exceptions,
such as NotAuthorized,
in the Opaque Error Types `obj` field,
and un-wrap them.

*Description of changes:*
- The Above
- Treat Markdown usage in documentation as warning instead of a danger; otherwise, current consumers (DB-ESDK) cannot pick up the latest Smithy-Dafny

Replaces https://github.com/smithy-lang/smithy-dafny/pull/464

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/ToDafnyAwsV2.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/ToDafnyAwsV2.java
@@ -689,11 +689,17 @@ public class ToDafnyAwsV2 extends ToDafny {
       .methodBuilder("Error")
       .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
       .returns(subject.dafnyNameResolver.abstractClassForError())
-      .addComment("While this is logically identical to the other Opaque Error case,")
+      .addComment(
+        "While this is logically identical to the other Opaque Error case,"
+      )
       .addComment("it is semantically distinct.")
-      .addComment("An un-modeled Service Error is different from a Java Heap Exhaustion error.")
+      .addComment(
+        "An un-modeled Service Error is different from a Java Heap Exhaustion error."
+      )
       .addComment("In the future, Smithy-Dafny MAY allow for this distinction.")
-      .addComment("Which would allow Dafny developers to treat the two differently.")
+      .addComment(
+        "Which would allow Dafny developers to treat the two differently."
+      )
       .addParameter(
         subject.nativeNameResolver.baseErrorForService(),
         "nativeValue"
@@ -713,11 +719,17 @@ public class ToDafnyAwsV2 extends ToDafny {
       .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
       .returns(subject.dafnyNameResolver.abstractClassForError())
       .addParameter(Exception.class, "nativeValue")
-      .addComment("While this is logically identical to the other Opaque Error case,")
+      .addComment(
+        "While this is logically identical to the other Opaque Error case,"
+      )
       .addComment("it is semantically distinct.")
-      .addComment("An un-modeled Service Error is different from a Java Heap Exhaustion error.")
+      .addComment(
+        "An un-modeled Service Error is different from a Java Heap Exhaustion error."
+      )
       .addComment("In the future, Smithy-Dafny MAY allow for this distinction.")
-      .addComment("Which would allow Dafny developers to treat the two differently.")
+      .addComment(
+        "Which would allow Dafny developers to treat the two differently."
+      )
       .addStatement(
         "return $T.create_Opaque(nativeValue)",
         subject.dafnyNameResolver.abstractClassForError()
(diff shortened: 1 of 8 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/ToDafnyAwsV2.java
```diff
@@ -727,11 +693,24 @@ MethodSpec generateConvertOpaqueError() {
         subject.nativeNameResolver.baseErrorForService(),
```
robin-aws, line 704: AFAICT the only difference between `generateConvertOpaqueServiceError` and `generateConvertOpaqueError` is the parameter type (the service's bass exception type vs. `Exception`). Couldn't we just have the latter in that case?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/ToNativeAwsV2.java
```diff
@@ -511,4 +512,66 @@ protected MethodSpec modeledUnion(final UnionShape shape) {
...
+    return initializeMethodSpec(methodName, inputType, returnType)
+      // If obj is an instance of the Service's Base Exception
+      .beginControlFlow(
+        "if ($L.$L instanceof $T)",
+        VAR_INPUT,
+        Dafny.datatypeDeconstructor("obj"),
+        subject.nativeNameResolver.baseErrorForService()
+      )
+      .addStatement(
+        "return ($T) $L.$L",
```
robin-aws, line 538: This doesn't end up behaving any differently from the next block AFAICT, since it only casts to a more specific exception type, but that distinction is lost since since the return type of the method is just `RuntimeException`.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/AwsSdkNativeV2.java
```diff
@@ -84,6 +85,13 @@ public AwsSdkNativeV2(final ServiceShape serviceShape, final Model model) {
         Map.entry("com.amazonaws.dynamodb", "DynamoDbException"),
         Map.entry("com.amazonaws.s3", "S3Exception")
       );
+    // These short names are used for convince in docs and errors messages.
```
robin-aws: How about using [`smithy.api#title`](https://smithy.io/2.0/spec/documentation-traits.html#smithy-api-title-trait) instead? That's defined for all relevant services, likely all AWS services in fact. And it should be the most human-readable and marketing-friendly name for a service, such as "Amazon Simple Storage Service" or "Amazon DynamoDB"

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/ToNativeAwsV2.java
```diff
@@ -511,4 +512,66 @@ protected MethodSpec modeledUnion(final UnionShape shape) {
...
+      )
+      // If String is set
+      .nextControlFlow(
+        "else if ($L.$L.$L())",
+        VAR_INPUT,
+        Dafny.datatypeDeconstructor("message"),
+        Dafny.datatypeConstructorIs("Some")
+      )
+      .addStatement(
+        "final $T message = $L($L.$L.$L)",
```
robin-aws: Perhaps this should stringify the opaque error value as well, like the last case below.

robin-aws (changes requested):
Definitely a big improvement to make sure a native error value is unwrapped so it round trips through Dafny, and it seems doable to ensure the same thing in any language (assuming we can safely hold onto such a value with a Dafny `object` reference, but that's an orthogonal issue).

It seems to be we can simplify a bit to avoid more cases that only work well for known services though - see other comments.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/ToDafnyAwsV2.java
texastony: While they behave identically today, they, ultimately, should not.
I will address via a comment.

Developers MAY treat un-modled KMS Exceptions differently than truly unexpected exceptions.
i.e: NotAuthorized vs Java Heap out of Memory 

That both of these exceptions end up as Opaque, today, is a miss.

It does not allow Dafny developers to treat the two separately.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/ToNativeAwsV2.java
texastony: We agreed to keep it.
I have documented why.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/AwsSdkNativeV2.java
texastony: Done.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/awssdk/v2/ToNativeAwsV2.java
texastony: Done.

seebees (approved):
Looks good.
I like adding the comments to the generated code. Though I do wish that there was a way to manage the comments and have a single string for them. However, I'm not sure how to do this cleanly and not overly complicate the code generation code...

robin-aws (approved):
🚀

PR #469, opened 2024-07-03
chore(tests): Add lite versions of DDB and KMS test models

### LINKED ISSUE #371: Add TestModel for simplified AWS SDK (SDKLite)
Probably best to make a filtered version of KMS. Goal is to reduce the size to make it an easier target for new language support.

robin-aws:
*Issue #, if available:*
Resolves #371 

*Description of changes:*
Adds copies of `ddb` and `kms` that project away all operations that aren't used by the MPL. Provides a much smaller beneration target for new language support.


TestModels/aws-sdks/ddb-lite/build.gradle.kts
```diff
@@ -0,0 +1,94 @@
...
+    implementation("software.amazon.smithy.dafny:smithy-dafny-codegen:0.1.0")
+}
+
+configure<software.amazon.smithy.gradle.SmithyExtension> {
+    // Uncomment this to use a custom projection when building the JAR.
+    // projection = "foo"
+}
+
+// Uncomment to disable creating a JAR.
+tasks["jar"].enabled = false
```
texastony: Question/Confusion: the Jar is disabled if this is uncommented.
So if you commented it, would that enable the Jar?

TestModels/aws-sdks/kms-lite/build.gradle.kts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+configure<software.amazon.smithy.gradle.SmithyExtension> {
+    // Uncomment this to use a custom projection when building the JAR.
+    // projection = "foo"
+}
+
+// Uncomment to disable creating a JAR.
+tasks["jar"].enabled = false
+
```
texastony: Also wrong or confusing.

TestModels/aws-sdks/kms-lite/smithy-build.json
```diff
@@ -0,0 +1,28 @@
+{
+  "version": "1.0",
+  "projections": {
+    "operation-subset": {
+      "transforms": [
+        {
+          "name": "excludeShapesBySelector",
+          "args": {
+            "selector": "operation :not([id|name = GenerateDataKey, Encrypt, Decrypt, GenerateDataKeyWithoutPlaintext, ReEncrypt])"
```
texastony: @josecorella do we need another Operation for ECDH?

Not that this is directly influences Smithy-Dafny or the correctness of this PR,
but I like some consistency;
I fear that we might copy and paste and fail...

TestModels/aws-sdks/kms-lite/test/TestComAmazonawsKms.dfy
texastony, line 1: Issue/Non-Blocking: Oh... this is third copy of this file to be checked in.
There is one in the MPL proper, one for the "heavy" test model, and one here.
Damn shame we have not figured out to reference identical files via Symbolic Links or some such.

TestModels/aws-sdks/kms-lite/smithy-build.json
robin-aws: I agree and the point of this list was to make what MPL needs, so I'd prefer to get it right even if it doesn't affect the validity of the PR.

TestModels/aws-sdks/ddb-lite/build.gradle.kts
robin-aws: Ah that's from the Smithy Gradle plugin template, agreed it's just confusing to leave it in when we never ship jars with just Smithy models in them around. Removed it from these models and `sqs` where it originally snuck in.

TestModels/aws-sdks/kms-lite/test/TestComAmazonawsKms.dfy
robin-aws: IKR, I had the same thought. `include`-ing a test file from a different test model would just pull in all the conflicting Dafny code. But I preferred to make sure these tests still worked on the projected model, and duplication across test models doesn't keep me up at night as much as many other things. :)

TestModels/aws-sdks/kms-lite/smithy-build.json
texastony: > DeriveSharedSecret and GetPublicKey are the ones used for ecdh

From an older Slack DM with @josecorella

TestModels/aws-sdks/kms-lite/model/model.json
seebees, line 1: I would use the 2.0 version from https://github.com/aws/aws-cryptographic-material-providers-library/blob/main/ComAmazonawsKms/Model/kms/model.json

seebees, line 1: I would use the 2.0 version from here: 

https://github.com/aws/aws-cryptographic-material-providers-library/blob/main/ComAmazonawsKms/Model/kms/model.json

TestModels/aws-sdks/kms-lite/smithy-build.json
seebees:
```suggestion
            "selector": "operation :not([id|name = GenerateDataKey, Encrypt, Decrypt, GenerateDataKeyWithoutPlaintext, ReEncrypt, DeriveSharedSecret, GetPublicKey])"
```

seebees (changes requested):
Can we add the 2 operations and update the Smithy version?

seebees (dismissed):
LGTM

seebees (approved):
LGTM

PR #470, opened 2024-07-03
fix(Java-SDK): dtor_message is never safe for Opaque

texastony:
*Issue #, if available:*

> Dafny’s compiled Java is very dangerous to use outside of Dafny compiled Java; the best advice we can give is to always write code as if you were the Dafny Compiler. For example, when working with a Datatype, only call a type’s deconstructor’s member if you KNOW the type is that deconstructor. Dafny’s compiled Java does NOT make such operations safe in any way.

The issue here is that we are calling a member of a deconstructor on an instance we KNOW is NOT a deconstructor that has the member. 

As such, calling `dtor_message` on `Error_Opaque` would lead to a Cast Exception. 

The fix is to not call it. 

*Description of changes:*
- Remove `dtor_message` block from Opaque_Error ToNative conversion method


robin-aws (approved):
Better to merge than not given we know this breaks downstream. But why didn't the Error TestModel catch this?

texastony:
> Better to merge than not given we know this breaks downstream. But why didn't the Error TestModel catch this?

I encourage a new issue tracking that.
I have a thesis: the branch is never used. 

It would only ever be used if a custom implementation of a KMS Keyring was passed in.

PR #478, opened 2024-07-16
chore(Makefile): update make targets to new cli

josecorella:
*Issue #, if available:*

*Description of changes:*


check_dafny_version.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
lucasmcdonald3: Put this in `scripts/` maybe?

SmithyDafnyMakefile.mk
```diff
@@ -82,47 +82,45 @@ ENABLE_EXTERN_PROCESSING?=
 # Verify the entire project
 verify:Z3_PROCESSES=$(shell echo $$(( $(CORES) >= 3 ? 2 : 1 )))
 verify:DAFNY_PROCESSES=$(shell echo $$(( ($(CORES) - 1 ) / ($(CORES) >= 3 ? 2 : 1))))
+verify:DAFNY_OPTIONS=--allow-warnings
```
lucasmcdonald3, line 87: Curious, what's the reasoning to take `allow-warnings` out of the commands and put it as DAFNY_OPTIONS? It looks like you added it back everywhere?

SmithyDafnyMakefile.mk
```diff
@@ -82,47 +82,45 @@ ENABLE_EXTERN_PROCESSING?=
...
+		--resource-limit $(MAX_RESOURCE_COUNT) \
 		$(DAFNY_OPTIONS) \
 		%
 
 # Verify single file FILE with text logger.
 # This is useful for debugging resource count usage within a file.
 # Use PROC to further scope the verification
+verify_single:DAFNY_OPTIONS=--allow-warnings
 verify_single:
 	dafny \
```
robin-aws: This is actually broken because you're missing `verify` (obviously never tested in CI :)

SmithyDafnyMakefile.mk
```diff
@@ -82,47 +82,45 @@ ENABLE_EXTERN_PROCESSING?=
...
+		--resource-limit $(MAX_RESOURCE_COUNT) \
 		$(DAFNY_OPTIONS) \
 		$(if ${PROC},-proc:*$(PROC)*,) \
 		$(FILE)
 
 #Verify only a specific namespace at env var $(SERVICE)
+verify_service:DAFNY_OPTIONS=--allow-warnings
 verify_service:
 	@: $(if ${SERVICE},,$(error You must pass the SERVICE to generate for));
 	dafny \
```
robin-aws: Likewise

SmithyDafnyMakefile.mk
```diff
@@ -151,20 +149,16 @@ dafny-reportgenerator:
 clean-dafny-report:
 	rm TestResults/*.csv
 
+check_dafny_version:
```
robin-aws: nit, but I'd name this `get_dafny_version` since it's not just a check, it actually sets DAFNY_VERSION which is read by other targets.

And once this is merged we should broadcast that you don't have to set DAFNY_VERSION yourself any more!

SmithyDafnyMakefile.mk
josecorella: ahh i forgot to add a meta comment about this.
--allow-warnings shouldn't be part of the overall dafny transpile/verify commands since we want to remove it sometime soon. Making it an option even though it is still used everywhere made sense to me

.github/workflows/pull.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     steps:
       - name: Populate Dafny versions list
         id: populate-dafny-versions-list
-        run: echo "dafny-versions-list=['4.2.0', '4.4.0', '4.8.0']" >> $GITHUB_OUTPUT
+        run: echo "dafny-versions-list=['4.8.0']" >> $GITHUB_OUTPUT
```
robin-aws: If we're claiming 4.5 and higher works, we should still test on 4.5. If not then get_dafny_version.sh should just assert the version is 4.8 or higher.

robin-aws (changes requested):
Great stuff, just some quick fixes

robin-aws (approved):

PR #483, opened 2024-07-22
Python: Makefile changes

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*


seebees (commented):
Looks good to me. Do we know that this works with our project? (Like the MPL?) (Then I can approve)

lucasmcdonald3:
Yes, I build the MPL with this.

seebees (approved):
Looks good, this unblocks and is not going into main yet.

josecorella (approved):
as a heads up, there is a known stack overflow problem when using verification when using the new dafny cli with 4.7.0 but that shouldn't block you and as ryan says this is not landing in main yet.
https://t.corp.amazon.com/P141728196/overview

PR #485, opened 2024-07-23
chore(TestModels): Add constraints inside list, map and union

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*
This PR adds constraints inside list (`ListWithConstraint`), map (`MapWithConstraint`) and union (`UnionWithConstraint`) in the test model.


TestModels/Constraints/Model/Constraints.smithy
```diff
@@ -96,6 +96,11 @@ list ListLessThanOrEqualToTen {
   member: String
 }
 
+@length(min: 1, max: 10)
+list ListWithConstraint {
+  member: MyString
```
rishav-karanjit, line 101: Self commenting a note posterity:
MyString has constraint `@length(min: 1, max: 10)`

Other aggregate shapes that was added also has similar constraint

alex-chew:
It would be good to also add a constraint within multiple levels of structures, to ensure that we're not only validating input structures. For example

```
structure GetConstraintsInput {
  ... // existing members omitted for brevity
  inner: InnerStructure,
}

structure InnerStructure {
  str: NonEmptyString,
}
```

PR #486, opened 2024-07-23
Python: Misc build/project config changes

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*


.github/workflows/test_models_python_tests.yml
```diff
@@ -0,0 +1,129 @@
...
+
+      - name: Setup Dafny
+        uses: dafny-lang/setup-dafny-action@v1.7.0
+        with:
+          dafny-version: ${{ matrix.dafny-version }}
+
+      - name: Setup Python for running tests
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.11
```
kessplas, line 97: nit: do we want to test 3.12 as well?

kessplas (approved):
approved with 1 nit

.github/workflows/test_models_python_tests.yml
lucasmcdonald3: I want to, but it's not really worth the effort/extra time to run the tests imo, especially since we just test 1 version on NET/Java

PR #492, opened 2024-07-25
Python: Most TestModels

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/TestModels/Aggregate/runtimes/python/pyproject.toml
+++ b/TestModels/Aggregate/runtimes/python/pyproject.toml
@@ -1,0 +1,26 @@
+[tool.poetry]
+name = "simple-aggregate"
+version = "0.1.0"
+description = ""
+authors = ["AWS <aws-arg-dafny@amazon.com>"]
+packages = [
+    { include = "simple_aggregate", from = "src" }
+]
+# Include all of the following .gitignored files in package distributions,
+# even though it is not included in version control
+include = ["**/smithygenerated/**/*.py", "**/internaldafny/generated/*.py"]
+
+[tool.poetry.dependencies]
+python = "^3.11.0"
+DafnyRuntimePython = "^4.4.0"
+# TODO: Depend on PyPi once Smithy-Python publishes their Python package there
+smithy-python = { path = "../../../../codegen/smithy-dafny-codegen-modules/smithy-python/python-packages/smithy-python", develop = false}
+
+standard-library = {path = "../../../dafny-dependencies/StandardLibrary/runtimes/python", develop = false}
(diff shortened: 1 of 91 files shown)


TestModels/Aggregate/runtimes/python/pyproject.toml
```diff
@@ -0,0 +1,26 @@
...
+    { include = "simple_aggregate", from = "src" }
+]
+# Include all of the following .gitignored files in package distributions,
+# even though it is not included in version control
+include = ["**/smithygenerated/**/*.py", "**/internaldafny/generated/*.py"]
+
+[tool.poetry.dependencies]
+python = "^3.11.0"
+DafnyRuntimePython = "^4.4.0"
+# TODO: Depend on PyPi once Smithy-Python publishes their Python package there
```
kessplas, line 16: nit: I suppose these comments aren't strictly accurate anymore? If I recall correctly, we plan to maintain a Smithy-Python fork indefinitely?

TestModels/Errors/src/WrappedSimpleErrorsImpl.dfy
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 include "../Model/SimpleErrorsTypesWrapped.dfy"
 
-module {:extern "simple.errors.internaldafny.wrapped"} WrappedSimpleErrorsService refines WrappedAbstractSimpleErrorsService {
```
kessplas, line 5: nit/question: did we decide how we're resolving the lack of extern trait for non-Python (and Go?) runtimes in 4.7?

kessplas (approved):
2 non-blocking comments, otherwise LGTM

TestModels/Aggregate/runtimes/python/pyproject.toml
lucasmcdonald3: Hopefully we don't maintain the fork indefinitely.

This comment is still accurate; if Smithy-Python were to publish their runtime to PyPI we should be able to depend on it immediately

TestModels/Errors/src/WrappedSimpleErrorsImpl.dfy
lucasmcdonald3: José is actively looking into this. I think the current plan is something like
```
module {:extern } WrappedSimpleErrorsService
```
but we're going to work together to see what makes sense over the next week or so. 

My plan is to have him resolve this on main for Java/NET, then I will pull in whatever the solution is.

PR #495, opened 2024-07-25
Python: AWS SDK TestModels

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*


TestModels/aws-sdks/sqs/build.gradle.kts
```diff
@@ -17,6 +17,7 @@ dependencies {
 
     // Must be built and published to the local Maven repo
     implementation("software.amazon.smithy.dafny:smithy-dafny-codegen:0.1.0")
+    implementation("software.amazon.smithy.python:smithy-python-codegen:0.1.0")
```
kessplas: question: do you plan on adding `sqs` to Python? 
opinion: this test model uses the Gradle build plugin, so it might be nice to have coverage, if that's supported in Python.

kessplas (approved):
one question, but LGTM

TestModels/aws-sdks/sqs/build.gradle.kts
lucasmcdonald3: We'll do Python SQS at some point, but it'll be at the same time as other languages.
(Probably when we start integrating the JSONEncryptor into SDK products)

I think I just needed this to fix the GHA build for this...
I'm not sure though.
I'm going to uncommit this and will recommit if I actually need this.

lucasmcdonald3:
```suggestion
```

lucasmcdonald3: Removed this.
I don't need reapproval to merge into the non-main branch, so I'll just merge it in.
Let me know if you want to talk about the removed line, but I'm assuming this is fine.

Issue #278, opened 2023-07-08
Generated Java cannot handle Optional Fields properly


texastony:
#### Issue
A Smithy Structure with an Optional Number member is represented in Dafny Source as
`Option<int32>`.

In Smithy Generated Java, the type for this member on the Structure's class will be `int`. 

`null` cannot be assigned to a Java `int`. 

As such, 
`Wrappers_Compile.Option.None` cannot be represented by 
the Java type.

#### Suggested Fix V2.1
To avoid a breaking change AND resolve this issue,
the only fix I can think of is to follow the pattern established by .NET implementation.

The "set-state" of a member MUST be tracked.
The "set-state" is a boolean flag that defaults to `false`.
The "set-state" switches to `true` when a value is provided for the member
during a Structure Shape's construction.
When a Structure is validated,
if the "set-state" of a  `@required` member is `false`,
the validation MUST fail.

In Java, 
this means the Builders of Structures
needs to be refactored to generate and utilize
"set-state" flags for each member.

The "set-state" flags need to be included on the final "built" structure.

In the ToDafny and ToNative classes,
these "set-state" flags can then be used to determine if no value was set.

#### Breaking Change Fix that is better.
Avoid Java Primitive types all together.
Or at least for non-required members.

This is going to complicate the type conversion.

Dafny Generated Java represents Optional Members as `Wrappers_Compile.Option<primitive>`.
Thus, we will be converting from non-primitive types (`Integer`) to primitive types (`int`).

An alternative approach would be to use Java's `Optional`.

However, Java discourages using `Optional` as a parameter or field;
it SHOULD BE exclusively used a Return type.

robin-aws:
The ideal pattern would be to follow what the current Java v2 SDKs do: required members use primitives like `int` directly, optional members use `Integer` so they can be `null`. E.g. https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/sqs/model/SendMessageRequest.html#delaySeconds()

Edit: it's possible even required members use `Integer` as well, I didn't immediately find one I know is required to confirm.

texastony:
I searched around, and could not find AWS SDK V2 primitives. 

From a "parsing customer input" perspective, I can see the benefit of having nullable fields. 

It affords differentiation between a customer setting/providing a value and not. 

I'm surprised I cannot find an opinion on this in the V2 Smithy Spec. 

I found it in the V1 Spec.

robin-aws:
[...]

PR #496, opened 2024-07-26
fix: use nullable types for all structure members (Java, .NET)

alex-chew:
*Issue #, if available:* #278

*Description of changes:* Changes the .NET and Java code generators to use nullable types for all structure members (even required members with primitive-equivalent shapes).


alex-chew:
Marking this as ready for review despite the unrelated failing CI test (something to do with a Rust-related patch not applying, which we believe is due to unintended nondeterminism in the module order emitted by the Dafny-to-Rust compiler).

robin-aws (changes requested):
Leaving an initial blocking review just because I'm not worried that this isn't correct for Smithy 1.0 models, that the better fix is to support @box consistently (it is supported for Java but not .NET AFAICT)

texastony:
An issue with going to down further down this path is that Smithy-Dafny is forking the Smithy 1.0 Specification.

Smithy-Dafny could argue that such a change in behavior is necessary to support the Local Service Trait; 
but I am not certain that argument is valid,
as I think the Local Service Trait can work without this forking of behavior.

A stronger argument, in favor of the box approach, 
or maybe just in dropping Smithy 1.0 and upgrading to 2.0,
is that all Smithy Code Generators should behave same by following the Smithy Specification.

robin-aws:
> An issue with going to down further down this path is that Smithy-Dafny is forking the Smithy 1.0 Specification.
> 
> Smithy-Dafny could argue that such a change in behavior is necessary to support the Local Service Trait; but I am not certain that argument is valid, as I think the Local Service Trait can work without this forking of behavior.
> 
> A stronger argument, in favor of the box approach, or maybe just in dropping Smithy 1.0 and upgrading to 2.0, is that all Smithy Code Generators should behave same by following the Smithy Specification.

I agree this deviation isn't related to local services. I suspect it's just that we misinterpreted the Smithy 1.0 semantics everywhere but Java.

Tactically though, given all the existing smithy-dafny models are currently relying on this misinterpretation, I say it's better to align Java and .NET with the models' intent and how they work in the other languages. We should document this explicitly as a smithy-dafny difference for now as part of this PR as well. 

Beyond that, we should ensure smithy-dafny DOES align with Smithy semantics on Smithy 2.0 models.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/Native.java
```diff
@@ -144,21 +143,14 @@ public TypeName typeForShape(final ShapeId shapeId) {
...
-        yield shape.hasTrait(BoxTrait.class) ? typeName.box() : typeName;
-      }
+      // All primitives are boxed for safety, even if @required
+      case BOOLEAN,
+        BYTE,
+        SHORT,
+        INTEGER,
+        LONG,
+        FLOAT,
+        DOUBLE -> NATIVE_TYPES_BY_SIMPLE_SHAPE_TYPE.get(shape.getType());
```
texastony, line 153: I knew I had written logic to handle this!
If I only Smithy-2.0 had been published earlier... 
I literally quoted the Smithy-1.0 spec...

robin-aws (approved):
My only comment was going to be that a bit more testing would be good, but then I recalled that I had to hamstring my smokeTests test on Constraints due to this issue. Now that it's fixed I was able to push a commit to remove the workaround, which addresses my concern about testing coverage, so I'm happy! :)

PR #504, opened 2024-08-06
feat: Support more services on Java

robin-aws:
*Description of changes:*
Tweaking several aspects of Java codegen to generalize more effectively beyond DDB and KMS:

* Implement a general pattern for determining the base exception and client typesRE for a service based on the `sdkId`, rather than hardcoding only DDB or KMS.
* Removed the `shapeRequiresTypeConversionFromStringToStructure` special casing, as it didn't work correctly for other services, and it actually turns out that the KMS project in the MPL was already patching it back to the correct thing anyway :)
* Handling more name transformations generally:
  * Using `Utils.unCapitalize` from the `awssdk:codegen` package instead of our own copy, to match names better.
  * Adding `Request`/`Response` to the name of the operation, instead of always replacing `Input`/`Output` suffixes of any structure.
* Switched the lakeformation test model to omit the `GetWorkUnitResults` operation, since it has a streaming result. This is specifically a blocker on Java because the result type becomes a `ResponseInputStream<GetWorkUnitResultsResponse>` instead of a `GetWorkUnitResultsResponse`, so the shim code doesn't compile otherwise.

I verified that regenerating the DDB and KMS java code in MPL does not change on this version, aside from making a patching unnecessary as above.


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/AwsSdkNativeV2.java
```diff
@@ -162,7 +152,7 @@ private TypeName typeForShapeNoEnum(ShapeId shapeId) {
     final Shape shape = model.expectShape(shapeId);
 
     if (shape.hasTrait(EnumTrait.class)) {
-      if (shapeRequiresTypeConversionFromStringToStructure(shapeId)) {
+      if (true || shapeRequiresTypeConversionFromStringToStructure(shapeId)) {
```
texastony: An `or` with a constant true is illogical. 

Is there a TODO here? Or should we just remove the if?

robin-aws: Implicit TODO, I was experimenting with whether we needed this test anymore. Will resolve before marking the PR actually ready :)

texastony: Yeah... there was an edge case with Enums and the Java SDK
that I used this to work around.

I swear I described it somewhere in Java codegen...

SmithyDafnyMakefile.mk
```diff
@@ -525,6 +525,7 @@ transpile_test_java: _transpile_test_all _mv_test_java
 # To avoid `java/implementation-java` the code is generated and then moved.
 _mv_implementation_java:
 	rm -rf runtimes/java/src/main/dafny-generated
+	mkdir -p runtimes/java/src/main
```
texastony, line 528: Is this right?
If you run `mkdir` against an existing directory, does it succeed?
Or does it fail?

robin-aws: It succeeds - `-p` makes it recursively create all missing directories, so it's perfectly happy if all already exist

robin-aws: Notice the same thing is already on the `runtimes/java/src/test` directory. I actually don't know how we got away without this fix for so long. :)

texastony: Because you did not use to generate the template files,
and we had to manually stub the __default.java files in the Java extern source path.

Thus, we always had a directory.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/Native.java
```diff
@@ -378,4 +372,8 @@ public ClassName baseErrorForService() {
       "Use `ClassName.get(RuntimeException.class)` instead."
     );
   }
+
+  protected ClassName classNameForAwsSdkShape(final Shape shape) {
+    throw new UnsupportedOperationException();
```
alex-chew: Can we add a message here to explain why it's unsupported?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/AwsSdkNativeV2.java
```diff
@@ -96,26 +100,7 @@ private void checkForAwsServiceConstants() {
...
-      AWS_SERVICE_NAMESPACE_TO_CLIENT_INTERFACE.containsKey(namespace);
-    if (!knowClientInterface) {
-      throw new IllegalArgumentException(
-        "Polymorph does not know this service's Client Interface: %s".formatted(
-            namespace
-          )
-      );
-    }
-  }
+  private static void checkForAwsServiceConstants(String namespace) {}
```
alex-chew: Is there a reason to keep this?

robin-aws: I had initially thought so, since although I've removed all the checks necessary there could be some in the future. But on second thought, I think those kinds of checks should be model validations instead anyway, so I've removed this.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/nameresolver/Native.java
robin-aws: Did you one better and made `Native` abstract so this could be abstract too. :)

robin-aws: Scratch that, the `Native` class is used directly for non SDK-style projects as well. I reverted and added an exception message instead. The better change would be to have an intermediate base class in between `Native` and `AwsSdkNativeV1/2`, but that looks like a substantial refactoring.

alex-chew (approved):

PR #508, opened 2024-08-08
Python: Timestamps for AWS SDKs

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*


TestModels/SimpleTypes/SimpleTimestamp/test/SimpleTimestampImplTest.dfy
```diff
@@ -19,7 +19,7 @@ module SimpleTimestampImplTest {
       modifies client.Modifies
       ensures client.ValidState()
     {
-        var dafnyTimestamp := "2024-06-11T12:34:56";
+        var dafnyTimestamp := "2024-06-11T12:34:56Z";
```
rishav-karanjit, line 22: Just wondering, what version of python does the customer have to be in because until python 3.10, it does not support military time stamp (timestamp with Z on the end)

lucasmcdonald3: 3.11. This constraint comes from Smithy-Python

lucasmcdonald3: I'll also note that I'm not even running SimpleTimestamp in Python, since I'm claiming that timestamps aren't supported for local services.

But the timestamp as-is is not ISO 8601 / RFC 3339 compatible, and there is complete consensus that any Dafny timestamps *should* be compatible with these specs.

rishav-karanjit (approved):

Issue #413, opened 2024-05-31
Benerate kms-lite for Rust


robin-aws:
Depends on #371

PR #510, opened 2024-08-12
feat: Partial support for SDK style libraries in Rust

robin-aws:
*Issue #, if available:*
#516

*Description of changes:*
First cut of actual codegen support for Rust, only for SDK-style libraries. I implemented this earlier in order to generate the "benerated" code in #413.

The code is not organized terribly well and definitely under-tested, but this will naturally improve as we replace beneration with actual code generation for all the TestModels.

`kms-lite` and `ddb-lite` are also updated with the latest generated code, and both include more operations that turned out to be used by the MPL and so on.

Known caveats:
* Unions are not supported yet - `ddb-lite/src/conversions/attribute_value.rs` is manually written.
* There is a lot of code duplication which we should clean up before considering Rust codegen complete. We should also factor out a lot of the conversion code generation as it will apply to local services as well.
* The Smithy build plugin doesn't support the same patching mechanism that the CLI does, so instead I've restored Mikael's hack to insert lines in `implementation_from_dafny.rs`, this time to declare the additional external modules. This will work for local services as well but I've only implement the SDK case for now.


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustAwsSdkShimGenerator.java
```diff
@@ -0,0 +1,1739 @@
...
+      "inputShapeName",
+      inputShapeName,
+      "outputType",
+      outputType,
+      "sdkId",
+      sdkId.toLowerCase(),
+      "clientName",
+      clientName,
+      "dafnyModuleName",
+      dafnyModuleName,
```
alex-chew, line 224: nit: Not at all blocking, but some of these variables seem to be unused (such as `clientName` and `dafnyModuleName`).

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustAwsSdkShimGenerator.java
```diff
@@ -0,0 +1,1739 @@
...
+    );
+
+    // Dafny maps smithy.api#Unit to ()
+    if (outputShape.getId() == ShapeId.from("smithy.api#Unit")) {
+      return TokenTree.of(
+        evalTemplate(
+          """
+          #[allow(dead_code)]
+          pub fn to_dafny(
+              value: &$sdkCrate:L::operation::$snakeCaseOperationName:L::$operationName:LOutput
```
alex-chew: nit: If we're special casing unit conversion, we may as well avoid an unused variable warning.
```suggestion
              _value: &$sdkCrate:L::operation::$snakeCaseOperationName:L::$operationName:LOutput
```

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustAwsSdkShimGenerator.java
```diff
@@ -0,0 +1,1739 @@
...
+  }
+
+  /**
+   * @param isRustOption Whether the Rust type is an Option<...> or not.
+   *                     Rust's structures avoid Options when not strictly necessary depending on context.
+   * @param isDafnyOption Whether the Dafny type is an Option<...> or not.
+   *                      Dafny tends to be much more consistent about this.
+   */
+  // TODO: There is obviously a lot of duplication here that should be easy to clean up.
+  // TODO: Some cases do not handle all combinations of isru
```
alex-chew: Is this supposed to say something like the following?
```suggestion
  // TODO: Some cases do not handle all combinations of isRustOption and isDafnyOption.
```

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustAwsSdkShimGenerator.java
```diff
@@ -0,0 +1,1739 @@
...
+  ) {
+    return switch (shape.getType()) {
+      case STRING -> {
+        if (shape.hasTrait(EnumTrait.class)) {
+          var enumShapeName = toSnakeCase(shape.toShapeId().getName());
+          if (isDafnyOption) {
+            yield TokenTree.of(
+              """
+              ::std::rc::Rc::new(match &%s {
+                  Some(x) => crate::_Wrappers_Compile::Option::Some { value: crate::conversions::%s::to_dafny(x.clone()) },
```
alex-chew, line 985: It seems unfortunate that we `.clone()` so much data that's about to be converted anyway (so presumably the clone would never live long), though I suppose performance isn't the goal of this first pass.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustAwsSdkShimGenerator.java
```diff
@@ -0,0 +1,1739 @@
...
+      .lineSeparated();
+    final var postlude = TokenTree.of(
+      """
+
+              // TODO: This should not be a panic, but the Dafny image of the enum shape doesn't have an Unknown variant of any kind,
+              // so there's no way to succeed.
+              // See https://github.com/smithy-lang/smithy-dafny/issues/476.
+              // This could be handled more cleanly if conversion functions returned Results,
+              // but that would be a large and disruptive change to the overall code flow.
+              _ => panic!("Unknown enum variant: {}", value),
```
alex-chew: Is there a reason to put this TODO in the generated code instead of in the generating code?

alex-chew (commented):
There's lots of room for refactoring and cleanup as you mentioned, but it's a huge milestone to finally have this part of the library code being generated.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustAwsSdkShimGenerator.java
robin-aws: Yup, performance definitely isn't the highest priority, as long as we can come back and improve this later. I definitely found cloning everywhere was much simpler, to avoid any interference between the different branches when converting a structure for e.g.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustAwsSdkShimGenerator.java
robin-aws: Probably! I don't think it's worth worrying about, and will be less of an issue if/when we switch to a `SimpleCodeWriter` where we just define variables as needed.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustAwsSdkShimGenerator.java
robin-aws: Good point, I'll move it to the code generator instead.

alex-chew (approved):

PR #511, opened 2024-08-12
[DRAFT] WIP "safe sed" for Python/Go/Rust

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

Tested with 2 TestModels (Boolean + Blob) and the MPL (all subprojects + test vectors).


ShubhamChaturvedi7 (approved):
LGTM from Go perspective. I’ve been just dropping the . from the externs so this change would require some refactor but this is what we want long term.

rishav-karanjit:
> LGTM from Go perspective. I’ve been just dropping the . from the externs so this change would require some refactor but this is what we want long term.

@ShubhamChaturvedi7 And I have been reverting your dropped `.` lol

lucasmcdonald3:
Closing in favor of https://github.com/smithy-lang/smithy-dafny/pull/512
Let's get this into main. Python can pull this in from main.

PR #512, opened 2024-08-13
chore: Optionally remove extern names when compiling to certain languages

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

* Script that does sed with lots of safety checks
* Makefile targets that plug into sed script
* Per-project Makefile string variables that are used by Makefile targets to remove extern strings

This is disabled by default and must be enabled per-project with the variable.


TestModels/SimpleTypes/SimpleBlob/Makefile
```diff
@@ -17,4 +17,18 @@ SERVICE_DEPS_SimpleBlob :=
 SMITHY_DEPS=dafny-dependencies/Model/traits.smithy
 
 # This project has no dependencies 
-# DEPENDENT-MODELS:= 
+# DEPENDENT-MODELS:=
+
+# Constants for languages that drop extern names (Python, Go)
```
robin-aws, line 25: Will all of these be required? That would really hurt smithy-dafny UX pretty badly.

SmithyDafnySedMakefile.mk
```diff
@@ -0,0 +1,104 @@
...
+_sed_types_file_add_extern:
+	$(MAKE) _sed_file SED_FILE_PATH=$(TYPES_FILE_PATH) SED_BEFORE_STRING=$(TYPES_FILE_WITHOUT_EXTERN_STRING) SED_AFTER_STRING=$(TYPES_FILE_WITH_EXTERN_STRING)
+
+_sed_index_file_add_extern:
+	$(MAKE) _sed_file SED_FILE_PATH=$(INDEX_FILE_PATH) SED_BEFORE_STRING=$(INDEX_FILE_WITHOUT_EXTERN_STRING) SED_AFTER_STRING=$(INDEX_FILE_WITH_EXTERN_STRING)
+
+_sed_wrapped_types_file_add_extern:
+	$(if $(strip $(WRAPPED_INDEX_FILE_PATH)), $(MAKE) _sed_file SED_FILE_PATH=$(WRAPPED_INDEX_FILE_PATH) SED_BEFORE_STRING=$(WRAPPED_INDEX_FILE_WITHOUT_EXTERN_STRING) SED_AFTER_STRING=$(WRAPPED_INDEX_FILE_WITH_EXTERN_STRING), )
+
+_sed_file:
```
robin-aws, line 82: This should really be in a separate script file if possible

TestModels/SimpleTypes/SimpleBlob/Makefile
lucasmcdonald3: Yes, all are required.
Chatted offline, I think I alleviated your concerns, but let me know if you still have concerns.

SmithyDafnyMakefile.mk
```diff
@@ -56,6 +56,15 @@ SMITHY_MODEL_ROOT := $(LIBRARY_ROOT)/Model
...
+# This flag enables pre-processing on extern module names.
+# This pre-processing is required to compile to Python and Go.
+# This is disabled by default.
+# This should be enabled in each individual project's Makefile if that project compiles to Python or Go.
+# This can be enabled by setting this variable to any nonempty value (ex. true, 1)
+ENABLE_EXTERN_PROCESSING?=
+
 ########################## Dafny targets
 
 # TODO: This target will not work for projects that use `replaceable`
```
robin-aws: As discussed offline, mind removing this comment since we've determined it's stale?

SmithyDafnySedMakefile.mk
```diff
@@ -0,0 +1,82 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# These are make targets that use sed to replace extern names.
+# Some Smithy-Dafny modules declare extern attributes with names that do not work in all languages.
+# These targets can remove or add those extern names based on the target language.
+# A target language SHOULD declare use of these targets.
+
+# These should eventually be replaced by replaceable modules that declare per-langauge extern names.
+# Eventually, the replaceable modules should be removed (or, their extern names should be removed),
+# and should be replaced by per-language package/namespace prefixes.
```
robin-aws, line 12:
```suggestion
# and should be replaced by per-language package/namespace prefixes.
See https://github.com/smithy-lang/smithy-dafny/issues/528.
```

TestModels/MultipleModels/Makefile
robin-aws, line 1: *Weeps softly*

robin-aws (commented):
As clean as such a solution can be. Just a couple of easy suggestions.

robin-aws:
Also just to check my understanding: for existing libraries only compiling to C# and/or Java, even setting `ENABLE_EXTERN_PROCESSING` is still ultimately a no-op, correct? Since the extra tasks just redundantly try to add the already-present externs?

lucasmcdonald3:
> Also just to check my understanding: for existing libraries only compiling to C# and/or Java, even setting ENABLE_EXTERN_PROCESSING is still ultimately a no-op, correct? Since the extra tasks just redundantly try to add the already-present externs?

No-op in the sense that "generated code is not expected to change" and "C#/NET will compile with the same Dafny code as before", yes.
But once you set it, you do still need to define the new variables. If `ENABLE_EXTERN_PROCESSING` is set without the new variables, things will fail.

TestModels/MultipleModels/Makefile
lucasmcdonald3: Followed up on https://github.com/smithy-lang/smithy-dafny/issues/528
I think this was another issue with replaceable modules; they didn't play nicely with our Makefile scripts to build "MultipleModels" projects.

SmithyDafnySedMakefile.mk
lucasmcdonald3: I put this in a slightly different place for clarity on what is actually addressed by #528; I propose that #528 is just an alternative interim solution

robin-aws (approved):
"A good compromise leaves everyone angry." :)

PR #514, opened 2024-08-14
Update to latest Smithy and remove workaround

robin-aws:
Since newer versions of Smithy reject the invalid `@smithy.rules#endpointRuleSet` trait on the SQS model, also removed that (attempted to update the SQS model but it having V2 enum shapes in it is still a blocker)


texastony (approved):
LGTM

PR #521, opened 2024-08-16
feat: Support externs in rust without patch files

robin-aws:
*Description of changes:*

Dafny's `feat-rust` branch currently assumes all code is going into one crate. There are two options for adding additional Rust code that is manually-written (or generated from other sources :):

1. Pass the files as additional input files to `dafny`.
2. Post-process the output from `dafny` to declare the extra modules.

#1 is cleaner, but requires that the Rust source files are not nested under folders to implicitly declare their ancestor modules. Since the design for Rust support so far has lots of these nested files, and we are hoping to eventually reuse the `smithy-rs` logic which generates such files, we're opting for #2 for now (although we do now use #1 for the StandardLibrary as a special case).

Our approach so far has been to use patch files to add these extra declarations, but maintaining the patch files has been very painful in development, and we don't want to force `smithy-dafny` consumers to use them to create functional Rust libraries. Therefore this change adds automation to add the extra declarations.

Since the extra declarations depend on the Smithy model, this is done in Java logic rather than Makefile or scripting. I've added an additional `patch-after-transpiling` command to the CLI, retroactively interpreting the existing single command as `generate`. The apache commons CLI library doesn't support this directly but it's pretty simple to handle subcommands ourselves.

Also deleted all patch files for Rust. 🎉  Now test models can indicate they are benerated with `RUST_BENERATED=1` instead, which will make `make polymorph_rust` a no-op. Also fixed up the test models to work with the common set of declarations: some test models were missing their copy of `standard_library_conversions.rs`, and `Constructor` deviated from the config naming convention.


TestModels/dafny-dependencies/dafny_runtime_rust/src/lib.rs
alex-chew, line 1: I didn't see these changes (or the others in `dafny_runtime_rust/src/*`) mentioned in the PR description - are they just copied from the latest feat-rust runtime?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
```diff
@@ -771,6 +775,85 @@ private void handlePatching(TargetLanguage targetLanguage, Path outputDir) {
     }
   }
 
+  public void patchAfterTranspiling() {
+    for (final TargetLanguage lang : targetLangOutputDirs.keySet()) {
+      switch (lang) {
+        case RUST -> patchRustAfterTranspiling();
+        default -> {}
```
alex-chew, line 782: Do you think it'd be worthwhile to emit a warning in case the subcommand gets accidentally used for other languages?

TestModels/dafny-dependencies/dafny_runtime_rust/src/lib.rs
robin-aws: Ah I had initially included updating the `feat-rust` commit in this PR too, but changed my mind. Reverting.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
robin-aws: I don't think so, since doing nothing is not wrong - there's just currently no need to patch any other languages.

alex-chew (approved):

PR #530, opened 2024-08-22
Python: Get most CI green

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

This gets most CI green. (Caveat: it does this by having separate transpile commands for NET/Java vs Python, but this SHOULD only be separate until NET/Java get on 4.8.)


.github/workflows/smithy-polymorph.yml
```diff
@@ -31,11 +31,17 @@ jobs:
...
           dafny-version: 4.1.0
 
-      - name: Install smithy-dafny-codegen dependencies locally
+      - name: Install smithy-dafny-codegen Rust dependencies locally
         uses: gradle/gradle-build-action@v2
         with:
           arguments: :codegen-client:pTML :codegen-core:pTML :rust-runtime:pTML
           build-root-directory: smithy-dafny-codegen-modules/smithy-rs
+      
+      - name: Install smithy-dafny-codegen Python dependencies locally
```
robin-aws: Just FYI I had to do this because `smithy-rs` packages aren't published at all yet, but I assume you have to do it here because you're still patching the `smithy-python` code?

robin-aws: Also it might be worth a reusable `.github/actions` for this since it's not very DRY.

codegen/settings.gradle.kts
```diff
@@ -5,9 +5,6 @@ rootProject.name = "smithy-dafny"
 include(":smithy-dafny-codegen")
 include(":smithy-dafny-codegen-cli")
 include(":smithy-dafny-codegen-test")
-// TODO: Once Smithy-Python is published to Maven, and we do not rely on a fork, use that
```
robin-aws, line 8: This should probably be tracked in an issue for both Python and Rust

codegen/smithy-dafny-codegen-test/src/test/java/software/amazon/polymorph/smithypython/PythonTestModels.java
```diff
@@ -0,0 +1,47 @@
...
+    static {
+        DISABLED_TESTS.add("AggregateReferences");
+        DISABLED_TESTS.add("LanguageSpecificLogic");
+        DISABLED_TESTS.add("SimpleTypes/BigDecimal");
+        DISABLED_TESTS.add("SimpleTypes/BigInteger");
+        DISABLED_TESTS.add("SimpleTypes/SimpleByte");
+        DISABLED_TESTS.add("SimpleTypes/SimpleFloat");
+        DISABLED_TESTS.add("SimpleTypes/SimpleShort");
+        DISABLED_TESTS.add("SimpleTypes/SimpleTimestamp");
+        DISABLED_TESTS.add("aws-sdks/ddb-lite");
```
robin-aws, line 27: FYI the -lite models should be very easy to turn on as well, since the projection is handled before hitting any language-specific logic.

SmithyDafnyMakefile.mk
```diff
@@ -156,9 +156,16 @@ clean-dafny-report:
 # Transpile the entire project's impl
 # For each index file listed in the project Makefile's PROJECT_INDEX variable,
 #   append a `-library:TestModels/$(PROJECT_INDEX) to the transpiliation target
-_transpile_implementation_all: TRANSPILE_DEPENDENCIES=$(patsubst %, --library:$(PROJECT_ROOT)/%, $(PROJECT_INDEX))
+_transpile_implementation_all: TRANSPILE_DEPENDENCIES=$(patsubst %, -library:$(PROJECT_ROOT)/%, $(PROJECT_INDEX))
 _transpile_implementation_all: transpile_implementation
 
+# TODO-new-cli: As part of supporting Dafny 4.8+, Java and NET SHOULD migrate to the new Dafny CLI,
```
robin-aws, line 162: This feels slightly off and could be more precise - is it because the legacy CLI will eventually be fully deprecated? It seems more like other languages need features that are only in the new CLI.

SmithyDafnyMakefile.mk
robin-aws, line 1: The `_new_cli` duplication seems reasonable to me. Only question is whether this PR also ensures they are as well tested. It seems like it is since Python will only use them?

TestModels/Aggregate/src/Index.dfy
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 include "SimpleAggregateImpl.dfy"
 
-module SimpleAggregate refines AbstractSimpleAggregateService {
+module {:extern "simple.aggregate.internaldafny"} SimpleAggregate refines AbstractSimpleAggregateService {
```
robin-aws, line 5: Is this a breaking change for existing projects?

TestModels/CodegenPatches/runtimes/python/pyproject.toml
robin-aws, line 1: Consider implementing `--generate project-files` for Python in the future, so many libraries don't have to provide this file.

SmithyDafnyMakefile.mk
lucasmcdonald3: Yes, right now only Python.
My goal is to get this into main soon so Rishav and Shubham can onboard to a "shared" solution.

SmithyDafnyMakefile.mk
lucasmcdonald3: Yes, but the more immediate desire is that Java/NET/Python/Go have the same build command.
This reduces risk of accidental "forks" in one build process or the other.
(I know we will need some Makefile checking that says "only add `--compile-suffix:1` for Java/NET", but that is a small "fork".)

This is meant to assert that "we have a 4.8 upgrade coming up, these targets SHOULD be merged as part of that upgrade"

TestModels/Aggregate/src/Index.dfy
lucasmcdonald3: Nope! This is the current state of main-1.x.
I'm putting the extern names back in the python-reviewed branch so that when I merge this branch in, there is no diff between its Dafny source and main's.

TestModels/CodegenPatches/runtimes/python/pyproject.toml
lucasmcdonald3: Good to know, thanks. I didn't realize we had this

.github/workflows/smithy-polymorph.yml
lucasmcdonald3: Yes and yes (both because Smithy-Python isn't on Maven *and* because I have patches.)

codegen/settings.gradle.kts
lucasmcdonald3: You mean an issue in Smithy-Python/rs that says "Publish codegen to Maven"? I agree

.github/workflows/smithy-polymorph.yml
lucasmcdonald3: I'm looking at this and don't see a *ton* of value in abstracting this to a resuable GHA?
I might hold off on this for now unless you think it's blocking.

lucasmcdonald3: Made a GHA for "Install Smithy-Dafny codegen dependencies"

codegen/settings.gradle.kts
lucasmcdonald3: Python: https://github.com/smithy-lang/smithy-python/issues/296
Rust already had one: https://github.com/smithy-lang/smithy-rs/issues/421

codegen/smithy-dafny-codegen-test/src/test/java/software/amazon/polymorph/smithypython/PythonTestModels.java
lucasmcdonald3: Took a look, I'll leave these unimplemented for now.
Even though the models are "lite", they use the gradle build process, which seems to need extra work to get working.
I think Python and Go will need extra work to figure out how to pass in extra args for the "dependency-module-names" and "library-name" flags.

.github/workflows/smithy-polymorph.yml
robin-aws: FYI I snuck these targets in for human usage too: https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/Makefile#L13 :)

Probably good to add the python dependency there too, perhaps in this PR so we don't forget?

codegen/smithy-dafny-codegen-test/src/test/java/software/amazon/polymorph/smithypython/PythonTestModels.java
robin-aws: Fair enough, there's no real motivation to get those working now anyway - they are mainly useful during the development of new languages and you're already past that stage.

TestModels/Aggregate/src/Index.dfy
robin-aws: Ah I actually missed that this is not targeting `main-1.x` itself.

robin-aws (approved):

PR #539, opened 2024-08-23
chore(tests): Include @dafnyUTF8Bytes in SimpleString test model

### LINKED ISSUE #519: Include @dafnyUtf8Bytes in SimpleString test model
This is an important wrinkle for string support that affects type conversion pretty substantially.

`GetStringUTF8` seems like it was intended to include this trait, but it was left off. Instead it is implemented identically to `GetString`, and the only difference is that the test that calls it provides a string that uses `\u` escape sequences.

robin-aws:
*Issue #, if available:*
Resolves #519 

*Description of changes:*
* Corrected `GetStringUTF8` to use this trait
* Corrected the description of the existing test that incorrectly said the string was in UTF-8 (it's a Dafny string so it's implicitly UTF-16 instead)
* Added identically structured test of `GetStringUTF8`
* Updated benerated Rust code to make sure this case is handled too.


TestModels/SimpleTypes/SimpleString/src/SimpleStringImpl.dfy
```diff
@@ -30,10 +30,10 @@ module SimpleStringImpl refines AbstractSimpleTypesSmithyStringOperations  {
         var res := GetStringOutput(value := input.value);
         return Success(res);
     }
-    method GetStringUTF8 ( config: InternalConfig,  input: GetStringInput )
-    returns (output: Result<GetStringOutput, Error>) {
+    method GetStringUTF8 ( config: InternalConfig,  input: GetStringUTF8Input )
```
seebees, line 36: There should be a `GetStringUTF8KnownValue` endpoint as well.

This catches some very subtle possible bugs in the to and from that cancel out. We have this in most of the simple types. Given how Go handles strings, I think this would be a good addtion.

seebees (changes requested):
Add a KnownValue test for UTF8 bytes plz

seebees (approved):
LGTM

PR #540, opened 2024-08-23
Python: Update as_dict/from_dict

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

* Fix as_dict/from_dict for SDK clients (just pass the client through)
* Override Smithy-Python to write dict parameters correctly (snakecase, see https://peps.python.org/pep-0008/#function-and-variable-names)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/extensions/DafnyPythonLocalServiceStructureGenerator.java
```diff
@@ -411,6 +435,94 @@ protected void writeFromDict(boolean isError) {
     writer.write("");
   }
 
+  /**
+   * Override Smithy-Python writeAsDict to handle shapes with {@link ReferenceTrait}s.
+   * Most of this is lifted directly from Smithy-Python; the changed components are
+   * highlighted.
```
ShubhamChaturvedi7: The whole thing is highlighted if you meant that the diff would be highlighted (In that case, it's better to remove the highlighted word).

ShubhamChaturvedi7 (commented):
Is CI expected to fail?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/extensions/DafnyPythonLocalServiceStructureGenerator.java
lucasmcdonald3: Oh no, by "highlighted" I meant that I was calling out specific areas.
But it is unclear. Let me clarify

lucasmcdonald3:
CI doesn't need to pass for this PR, but it was easy to make it pass. I updated this branch with the latest python-reviewed branch, and now most CI is passing.

The final check will pass after https://github.com/smithy-lang/smithy-dafny/pull/541, which formats the file in this PR with the updated logic from this PR, so I'll merge #541 in after this PR goes in.

ShubhamChaturvedi7 (approved):
Checked with Lucas, prettier runs here https://github.com/smithy-lang/smithy-dafny/pull/541

PR #542, opened 2024-08-24
feat: Rust codegen for more SimpleTypes models

### LINKED ISSUE #375: Support all benerated SimpleTypes TestModels on Rust
This should be able to reuse a lot of the codegen for the SDK case added by #510, since the shape conversion logic should be basically identical, just used in the opposite direction. Only SimpleTypes test models that are currently being benerated and tested, are in scope.

It MAY be possible to reuse `smithy-rs` logic to generate the local service shapes in Rust, but it's also likely much trickier to get that working than just following the same approach as #510. I'd recommend being aware of it but not using it yet, but potentially reconsidering it as we tackle #402.

Acceptance criteria:
* All in-scope SimpleTypes TestModels build and test successfully using code generation and with empty patch files (since `make polymorph_rust` generates everything)
  * The exact code output is definitely not set in stone, but we should retain all benerated functionality.
* All in-scope SimpleTypes TestModels' Makefiles specify `TRANSPILE_TESTS_IN_RUST=1`, meaning they include the Dafny tests in scope and are executing them in CI.
  * This is opt-in because it was significant additional work to benerate the necessary wrapped clients. Only SimpleBoolean currently has this enabled, but the equivalent code should be generated for all test models.

alex-chew:
*Issue #, if available:* Resolves #375.

*Description of changes:* This PR will implement Rust library codegen for the remaining SimpleTypes models that are currently benerated:

- [x] SimpleBlob
- [x] SimpleDouble
- [x] SimpleEnum
- [x] SimpleEnumV2
- [x] SimpleInteger
- [x] SimpleLong
- [x] SimpleString
- [x] SimpleTimestamp


ajewellamz (approved):
LGTM

PR #560, opened 2024-09-04
chore(tests): Extend LocalService and support on Rust

### LINKED ISSUE #374: Add operations on local service references (add to LocalService)
The `@reference` trait can be applied to service shapes, which results in references to service clients. The `LocalService` test model should include operations that accept and return such references, since they also need to be converted between Dafny and native language representations. 

This often requires a language's code generator to generate a constructor for a client that accepts a Dafny client implementation, rather than a native config shape.

Original title: "Add TestModel for alternate constructors (AltConstructor)"
Original description:

Extend the Constructor TestModel to validate that a language's code generator generates a constructor for a client that accepts a Dafny client implementation, rather than a native config shape.

Context:
Right now, this TestModel only validates that an operation on the client can return values passed into the constructor.
But in some cases (notably wrapping/unwrapping a KMS client from within a local service), we must create a native client by passing a Dafny client implementation, and not by passing in a native client configuration.

### LINKED ISSUE #559: Support LocalService on Rust
Valuable once https://github.com/smithy-lang/smithy-dafny/issues/374 is completed

### LINKED ISSUE #564: Using tokio::runtime::Runtime::block_on to invoke async functions from Dafny code is not re-entrant
We're using this approach by necessity, because all Dafny-generated Rust code is sequential, but Dafny code needs to invoke operations on service clients. Rust AWS SDK clients already only expose async functions, and we're following suit with local service clients as well. Hence both Dafny-wrapped SDKs and "wrapped for testing" local service clients (i.e. wrapped/client.rs) have an embedded Tokio runtime in `current_thread` mode for making these calls.

The problem is that local service code might in turn try to invoke functions on other clients, at which point the Tokio runtime panics with "Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks."

https://github.com/smithy-lang/smithy-dafny/pull/560 reveals this by having local service accept its own client type and making a call on an operation, but this pattern will also happen on real production smithy-dafny libraries: a wrapped MPL client will try to invoke a KMS SDK operation and hit the same problem.

The alternative of calling `spawn` and then `block_on` on the resulting future doesn't work because Dafny values generally don't implement `Send`, and achieving that in the Dafny Rust code generator is a huge lift.

Note we also want to use a single static Tokio runtime instead of one per client, since they are expensive to create and hold onto 4 file handles each. But that seems to be an orthogonal improvement and this issue is more critical to figure out first.

robin-aws:
*Issue #, if available:*
Resolves #374. Resolves #559. Resolves #564.

*Description of changes:*
Extends LocalService to actually pass around client references as well, which increases testing coverage a fair bit. See #374 for details.

Then extended Rust codegen to handle this test model. It theoretically should handle wrapping and unwrapping local service clients from dependencies, but this is difficult to test directly since our Rust support in general requires everything in a single crate.

Also addressed #564 since it was initially revealed by this test model.

Also disabled testing this model on Python since it tickled a similar bug related to `async`: https://github.com/smithy-lang/smithy-dafny/issues/566


TestModels/LocalService/test/SimpleLocalServiceTest.dfy
```diff
@@ -8,14 +8,35 @@ module SimpleLocalServiceTest {
...
+    requires client.ValidState()
+    modifies client.Modifies
+    ensures client.ValidState()
+  {
+    // This is a flat out lie and directly contradicts client.ValidState().
+    // But otherwise it's not possible to call other client methods like this.
+    // Since this test model is only using self-reference in order to
+    // avoid extra dependencies, and self-reference isn't that useful a use case in reality,
+    // lying to get around the specification restrictions in a test is not terribly risky.
+    assume {:axiom} client.Modifies !! {client.History};
```
seebees: My position on tests in Dafny is that sometimes it is better to outright lie to verify the behavior you want to test.

This is because Dafny verification is a kind of test, and so we write the specifications so we don't need to test. But if we want to confirm the spec, we need some lies.

robin-aws: I'd much rather not lie since you can't actually reliably bound the lie in general - I'm only comfortable here because it's only test code etc.

To be clearer the actual problem is that the generated spec for `SelfReflection` includes `requires input.self.Modifies !! {History}`, which is impossible when `input.self` is the client itself.

The good news is I realized I could avoid the issue by using a second client - see latest commit.

alex-chew (approved):
LGTM

PR #562, opened 2024-09-05
feat: real code generation for Constructor in Rust

ajewellamz:
*Issue #, if available:*

*Description of changes:*


TestModels/Constructor/runtimes/rust/Cargo.toml
```diff
@@ -5,13 +5,20 @@ edition = "2021"
...
 
+[features]
+wrapped-client = []
+
 [dependencies]
 aws-smithy-runtime = {version = "1.6.0", features = ["client"] }
 aws-smithy-runtime-api = {version = "1.7.0", features = ["client"] }
 aws-smithy-types = "1.2.0"
 dafny_runtime = { path = "../../../dafny-dependencies/dafny_runtime_rust"}
+dafny_standard_library = { path = "../../../dafny-dependencies/StandardLibrary/runtimes/rust" }
```
robin-aws: This is actually useless and you should remove it - because we're having to cram everything into a single crate, the copy of the Dafny `Wrappers` types in StandardLibrary is distinct from the copy this model actually runs with, so you can't use any of the utilities. That's why codegen is barfing out a copy of `standard_library_conversions.rs` every time :(

TestModels/Constructor/runtimes/rust/src/client/get_constructor.rs
```diff
@@ -1,19 +1,20 @@
...
     /// - On success, responds with [`GetConstructorOutput`](crate::operation::get_constructor::GetConstructorOutput) with field(s):
-    ///   - [`internal_config_string(Option<String>)`](crate::operation::get_constructor::GetConstructorOutput::internal_config_string): (undocumented)
-    ///   - [`blob_value(Option<Blob>)`](crate::operation::get_constructor::GetConstructorOutput::blob_value): (undocumented)
-    ///   - [`boolean_value(Option<bool>)`](crate::operation::get_constructor::GetConstructorOutput::boolean_value): (undocumented)
-    ///   - [`string_value(Option<String>)`](crate::operation::get_constructor::GetConstructorOutput::string_value): (undocumented)
-    ///   - [`integer_value(Option<i32>)`](crate::operation::get_constructor::GetConstructorOutput::integer_value): (undocumented)
-    ///   - [`long_value(Option<i64>)`](crate::operation::get_constructor::GetConstructorOutput::long_value): (undocumented)
+    ///   - [`blob_value(Option<::aws_smithy_types::Blob>)`](crate::operation::get_constructor::GetConstructorOutput::blob_value): (undocumented)
```
robin-aws, line 10: Minor point (not in scope for this PR), but it might be a bit better to emit things in the original smithy model order rather than sorting them.

TestModels/Constructor/runtimes/rust/src/conversions/get_constructor/_get_constructor_output.rs
robin-aws, line 1: Love how much simpler this code got :)

TestModels/Constructor/runtimes/rust/src/operation/get_constructor/_get_constructor_input.rs
```diff
@@ -1,25 +1,27 @@
...
-    pub fn value(&self) -> ::std::option::Option<&str> {
-        self.value.as_deref()
-    }
+pub fn value(&self) -> &::std::option::Option<::std::string::String> {
+    &self.value
+}
 }
 impl GetConstructorInput {
-    /// Creates a new builder-style object to manufacture [`GetConstructorInput`](crate::operation::get_constructor::GetConstructorInput).
+    /// Creates a new builder-style object to manufacture [`GetConstructorInput`](crate::operation::operation::GetConstructorInput).
```
robin-aws: This looks like a regression - might be an unrelated bug in the existing codegen though. In fact the old version wasn't 100% right either.

Is there a linter that can check the validity of these symbols in comments?

TestModels/Constructor/runtimes/rust/tests/constructor_test.rs
robin-aws, line 1: Can we keep this file? Yes it's not as strictly necessary since we're running the actual Dafny tests instead, but it's really good to have manually-written Rust tests too, since it can catch issues with round tripping through Dafny -> Rust -> Dafny hiding bugs. Plus this is nice to see what interacting with the library actually looks like.

robin-aws (changes requested):
LGTM, love that the actual codegen change was so small. Just some minor fixups.

TestModels/Constructor/runtimes/rust/Cargo.toml
ajewellamz: We should make it a habit to run `cargo machete` before a checking in a Cargo.toml

robin-aws (approved):

Issue #534, opened 2024-08-22
Support Errors on Rust


robin-aws:
This specifically means the single `Error` type as benerated by the `Errors` TestModel. The other test models were not updated to use this style yet.

robin-aws:
Resolved by https://github.com/smithy-lang/smithy-dafny/pull/573 (but with #576 for followup)

PR #573, opened 2024-09-11
feat: Rust codegen for Errors test model

alex-chew:
*Issue #, if available:* #534


ajewellamz:
These things are easier to review if you check in the *.rs files. We can always delete them in a following PR.

Along similar lines, maybe this was the PR to delete the *.rs files from Constructor, rather than updating them.

I see that the build() functions return aws_smithy_types::error::operation::BuildError, which are then wrapped in an Opaque, and so any actual message is forever lost to the customer. Is there any chance the we could define an additional Error variant for that instead?

That said, it's probably best to call this done for now, and make enhancements later.

ajewellamz (approved):

TestModels/Resource/runtimes/rust/src/types/error.rs
```diff
@@ -0,0 +1,26 @@
...
+        message: ::std::string::String,
+    },
+    Opaque {
+        obj: ::dafny_runtime::Object<dyn ::std::any::Any>,
+    },
+}
+
+impl ::std::cmp::Eq for Error {}
+
+impl ::std::fmt::Display for Error {
```
robin-aws, line 20: Is this not what `#[derive(::std::fmt::Display)]` would generate anyway?

TestModels/Resource/runtimes/rust/src/operation/get_resources/builders.rs
```diff
@@ -43,17 +43,18 @@ impl GetResourcesFluentBuilder {
...
-            // Using unhandled since GetResources doesn't declare any validation,
-            // and smithy-rs seems to not generate a ValidationError case unless there is
-            // (but isn't that a backwards compatibility problem for output structures?)
-            // Vanilla smithy-rs uses SdkError::construction_failure,
-            // but we aren't using SdkError.
-            .map_err(crate::operation::get_resources::GetResourcesError::unhandled)?;
+            // Using Opaque since we don't have a validation-specific error yet.
+            // Operations' models don't declare their own validation error,
+            // and smithy-rs seems to not generate a ValidationError case unless there is.
+            // Vanilla smithy-rs uses SdkError::construction_failure, but we aren't using SdkError.
```
robin-aws, line 54: This comment should probably be a comment on the codegen code instead at this point.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustLibraryShimGenerator.java
```diff
@@ -388,6 +382,32 @@ private RustFile typesBuildersModule() {
...
   }
 
+  private RustFile typesErrorModule() {
+    final Map<String, String> variables = serviceVariables();
+    final String modeledErrorVariants = allErrorShapes()
+      .map(errorShape ->
+        IOUtils.evalTemplate(
+          """
+          $rustErrorName:L {
+              message: ::std::string::String,
```
robin-aws, line 392: This is fine but reminded me I should cut https://github.com/smithy-lang/smithy-dafny/issues/575

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/AbstractRustShimGenerator.java
```diff
@@ -76,10 +76,13 @@ protected Stream<OperationShape> allOperationShapes() {
   }
 
   protected Stream<StructureShape> allErrorShapes() {
```
robin-aws, line 78: FYI for full completeness this needs to also include all errors in any dependent services. But that's not possible to test since we can't build test models with dependencies yet. So this is just fine to hit the bar of supporting the `Errors` test model in this PR, but we will need to follow up with https://github.com/smithy-lang/smithy-dafny/issues/576.

robin-aws:
> I see that the build() functions return aws_smithy_types::error::operation::BuildError, which are then wrapped in an Opaque, and so any actual message is forever lost to the customer. Is there any chance the we could define an additional Error variant for that instead?

Good callout that I think we can address in the scope of supporting `Constraints` (since that will actually add validation errors).

Even so, the message isn't actually lost, it's just buried and won't be printed out by default. Do we need to shoot for the display content of any unexpected error being printed when printing an `Opaque`? I'm not sure that's feasible, nor do the other languages try to achieve that AFAICT.

robin-aws (approved):
Awesome to have this done. Made non-blocking comments - they can be addressed in a future PR if necessary.

PR #577, opened 2024-09-11
fix: polymorph_rust accidentally also invoking patch_after_transpile

robin-aws:
*Description of changes:*

Classic case of forgetting `break;` statements in my switch statement. But switching (ha!) to a switch expression makes it a non-issue.

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
@@ -885,10 +885,6 @@ public class CodegenEngine {
       final List<String> lines = Files.readAllLines(
         implementationFromDafnyPath
       );
-      // TODO fix root cause of duplicate extra declarations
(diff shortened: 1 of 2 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
alex-chew, line 1: Not blocking, but since we no longer rely on the

```
// (extra-declarations)
```

we may as well remove it from `extraDeclarationsForLocalService`.

alex-chew (approved):

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
robin-aws: Good point, although I'm actually more inclined to beef up the comment to document the fact that it didn't come from Dafny.

ajewellamz (approved):

PR #582, opened 2024-09-13
feat: Rust codegen for Constraints

### LINKED ISSUE #533: Support Constraints on Rust

alex-chew:
*Issue #, if available:* #533

*Description of changes:*
This PR implements Rust library codegen for input validation of supported constraint traits (`@required`, `@length`, `@range`).

Notes:
* Unlike in Java and C#, library users may directly construct the generated operation input structures in Rust (in order to aid in idiomatic usage, especially w.r.t. pattern matching, like the AWS SDK for Rust does) and are not restricted to using the builder pattern. Therefore, the builder can't take sole responsibility for validation; instead, validation happens whenever an operation is called.
* Unlike other target languages, Rust is limited in its ability to propagate arbitrary errors up the call stack unless the API is designed for doing so. Our models don't define a specific error shape for validation, so in order to represent validation errors in the generated `Error` enum, we generate a `ValidationError` variant that only exists on the Rust side. When converting to Dafny, we upcast the error into the `Opaque` variant; when converting from Dafny, we downcast it back after checking that doing so is safe using `dafny_runtime::is_object!`.
* By Polymorph's design, values crossing the boundary from Dafny-compiled code are assumed to be "valid", but the `WrappedSimpleConstraintsTest` intentionally breaks this assumption in order to test the constraint validation logic, forcing invalid values across the boundary by (ab)using `{:axiom}`. This would be okay except that the Dafny tests in the Constraints test model *also* pass across malformed UTF-8 for `@DafnyUtf8Bytes` strings (which is not a constraint trait). The Rust type conversion process does not allow for failure when converting values from Dafny (doing so would be a major refactor and probably some performance impact), so I've isolated the malformed UTF-8 test assertions and used a bit of `sed` to simply disable them for Rust specifically, while allowing the other languages to continue testing them.


TestModels/Positional/runtimes/rust/src/operation/get_resource.rs
```diff
@@ -10,13 +10,20 @@ impl GetResource {
...
+
     pub(crate) async fn send(
         client: &crate::client::Client,
         input: crate::operation::get_resource::GetResourceInput,
     ) -> ::std::result::Result<
         crate::operation::get_resource::GetResourceOutput,
         crate::types::error::Error,
     > {
+        if input.name.is_none() {
+    return ::std::result::Result::Err(::aws_smithy_types::error::operation::BuildError::missing_field(
```
ajewellamz, line 22: I feel like going through a aws_smithy_types::error is adding complexity without adding value.
Couldn't we just make a wrap_validation_err out of a String and leave it as that?

robin-aws: I think it's valuable: it makes it possible to programmatically handle validation errors, and I can see that being useful for middleware that handles translation of Smithy-modeled types, for example.

As long as displaying these errors gives a good user experience I say we keep them.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustLibraryShimGenerator.java
```diff
@@ -705,6 +713,135 @@ private RustFile operationOuterModule(final OperationShape operationShape) {
...
+            min.orElse(""),
+            max.map(val -> "=" + val).orElse("")
+          );
+      final var rangeDescription = describeMinMax(min, max);
+      return IOUtils.evalTemplate(
+        """
+        if matches!(input.$fieldName:L, Some(x) if %s) {
+            return ::std::result::Result::Err(::aws_smithy_types::error::operation::BuildError::invalid_field(
+                "$fieldName:L",
+                "$fieldName:L failed to satisfy constraint: Member must be %s",
```
robin-aws, line 768: We should avoid using smithy terms like "constraint" or "Member" in these kinds of error messages. We're free to use very Rust-specific terms and that will be easier to understand.

I'd actually suggest we just steal the exact phrasing `smithy-rs` uses for these cases. Just recall that you have to generate a server SDK instead of a client to see the constraint validation logic.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustLibraryShimGenerator.java
```diff
@@ -705,6 +713,135 @@ private RustFile operationOuterModule(final OperationShape operationShape) {
...
+    ) {
+      final var targetShape = model.expectShape(memberShape.getTarget());
+      final var len =
+        switch (targetShape.getType()) {
+          case BLOB -> "x.as_ref().len()";
+          case STRING -> targetShape.hasTrait(DafnyUtf8BytesTrait.class)
+            // scalar values
+            ? "x.chars().count()"
+            // bytes
+            : "x.len()";
```
robin-aws: This is technically wrong, but it looks like the other languages are also wrong according to the Smithy spec: https://github.com/smithy-lang/smithy-dafny/issues/610.

Would it be a massive pain to call something that calculates the number of UTF-16 code points instead, to at least be consistent with the other languages?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustLibraryShimGenerator.java
```diff
@@ -705,6 +713,135 @@ private RustFile operationOuterModule(final OperationShape operationShape) {
...
+    ) {
+      if (min.isPresent() && max.isPresent()) {
+        return "between %s and %s, inclusive".formatted(min.get(), max.get());
+      } else if (min.isPresent()) {
+        return "greater than or equal to %s".formatted(min.get());
+      } else {
+        if (max.isEmpty()) {
+          throw new IllegalArgumentException(
+            "At least one of max and min must be non-null"
+          );
```
robin-aws, line 841: Just FYI this should be impossible to hit, since the trait validation already checks this

codegen/smithy-dafny-codegen/src/main/resources/templates/runtimes/rust/types/error.rs
```diff
@@ -2,9 +2,10 @@
 pub enum Error {
     $modeledErrorVariants:L
     CollectionOfErrors {
-        list: ::std::vec::Vec<Error>,
+        list: ::std::vec::Vec<Self>,
```
robin-aws, line 5: I assume this is just a style thing and semantically equivalent?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/AbstractRustShimGenerator.java
```diff
@@ -463,7 +463,10 @@ protected TokenTree fromDafny(
               valueToRust = "(%s).unwrap()".formatted(valueToRust);
             }
           } else {
-            valueToRust = dafnyToRust.formatted(dafnyValue + ".as_ref()");
+            valueToRust =
+              dafnyToRust.formatted(
+                "::std::borrow::Borrow::borrow(%s)".formatted(dafnyValue)
+              );
```
robin-aws, line 469: For my own education, what's the difference?

TestModels/Constraints/runtimes/rust/src/types/error.rs
```diff
@@ -0,0 +1,67 @@
...
+#[derive(::std::clone::Clone, ::std::fmt::Debug, ::std::cmp::PartialEq)]
+pub enum Error {
+    SimpleConstraintsException {
+    message: ::std::string::String,
+},
+    CollectionOfErrors {
+        list: ::std::vec::Vec<Self>,
+        message: ::std::string::String,
+    },
+    ValidationError(ValidationError),
```
robin-aws, line 13: Could this be `ValidationError(BuildError)` instead?

robin-aws (changes requested):
Nice job solving some tricky constraints on this (:D)

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustLibraryShimGenerator.java
alex-chew: Those suggestions are contradictory, as I did steal the (almost exact) phrasing from `smithy-rs` server constraint validation logic: see [`LengthTraitValidationErrorMessage`](https://github.com/smithy-lang/smithy-rs/blob/6b42eb5ca00a2dc9c46562452e495a2ec2e43d0f/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/LengthTraitValidationErrorMessage.kt#L12) and [`RangeTraitValidationErrorMessage`](https://github.com/smithy-lang/smithy-rs/blob/6b42eb5ca00a2dc9c46562452e495a2ec2e43d0f/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RangeTraitValidationErrorMessage.kt#L12).

codegen/smithy-dafny-codegen/src/main/resources/templates/runtimes/rust/types/error.rs
alex-chew: Yes. I find it nicer to minimize the number of places where I need to think about whether I'm talking about the locally-defined `Error` enum or the `std::error::Error` trait.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/AbstractRustShimGenerator.java
alex-chew: https://doc.rust-lang.org/std/convert/trait.AsRef.html#relation-to-borrow

The relevant point here is that `Borrow` has a blanket impl for any `T` (whereas both `Borrow` and `AsRef` have blanket impls for any `&T` and `&mut T`).

TestModels/Constraints/runtimes/rust/src/types/error.rs
alex-chew: For implementing constraint trait validation only, yes, I believe this could just directly wrap a `BuildError`.

There's two levels of indirection here:

- This PR uses an  `Rc<dyn std::error::Error>` instead of `BuildError` directly, because I was not (and am not) sure that all validation errors will be `BuildError`s.
- In order to safely up-/down-cast the contents of the `Error::ValidationError` variant from the `dyn Any` in an arbitrary `dafny_runtime::Object`, it's best to define a unique type that we own - that is, the `ValidationError` struct (not the enum variant). If we defined `Enum::ValidationError(Rc<...>)` or `Enum::ValidationError<BuildError>`, then it seems possible for us to mishandle a `dafny_runtime::Object` containing an `Rc<?>` or `BuildError` from somewhere else.

In my mind, this doesn't make the library harder to use: the typical library user would mainly care whether or not the error variant is a `ValidationError` or something else. If they want the message or the underlying `dyn std::error::Error`, there's no need to manually traverse the indirections: they can use `to_string()` and `source()` as you would for any other `std::error::Error`.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustLibraryShimGenerator.java
alex-chew: That's true, I only include this as a defensive measure. (FWIW smithy-rs does the same, e.g. [`RangeTrait.rangeDescription`](https://github.com/smithy-lang/smithy-rs/blob/6b42eb5ca00a2dc9c46562452e495a2ec2e43d0f/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RangeTraitValidationErrorMessage.kt#L23))

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustLibraryShimGenerator.java
robin-aws: Ha, alright then, I'll redirect my constructive criticism towards smithy-rs :) Good on you for doing the right thing in copying them anyway!

robin-aws: Realized my original callout is still valid, and a point where polymorph is different from server SDKs, so I cut an issue for follow-up: https://github.com/smithy-lang/smithy-dafny/issues/611. Fine to leave this as is for now though.

TestModels/Constraints/runtimes/rust/src/types/error.rs
robin-aws: Fair points. What would it look like for an end-user to react differently to a validation error, and actually dig out the offending field from the wrapped `BuildError`?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustLibraryShimGenerator.java
alex-chew: It turns out that it's not too difficult to count UTF-16 code points instead, so I've implemented it in the latest commit.

TestModels/Constraints/runtimes/rust/src/types/error.rs
alex-chew: There isn't a clean way to dig out the name of the offending field right now. Neither [`BuildError`](https://github.com/smithy-lang/smithy-rs/blob/2f29209c1ecfd8a848dc0745dbce6b510c898d9a/rust-runtime/aws-smithy-types/src/error/operation.rs#L91) nor its inner [`BuildErrorKind`](https://github.com/smithy-lang/smithy-rs/blob/2f29209c1ecfd8a848dc0745dbce6b510c898d9a/rust-runtime/aws-smithy-types/src/error/operation.rs#L67) give public visibility to their members.

Did you have a particular use-case in mind for needing to programmatically retrieve the name of the invalid field, rather than showing the full error message?

robin-aws: I see what you mean - I had thought `BuildErrorKind` was structured rather than just a string to enable programmatic handling, but it looks more like the main purpose is just to display a nice error message string. 

Looks like we're meeting the bar of the existing `smithy-rs` behavior so I think this is the right call for now. We can potentially revisit and improve as part of #598

robin-aws (approved):
🚀 (within carefully specified limits of course :)

PR #623, opened 2024-10-08
fix(ToDafny-ByteBuffer): Do not Modify Input

texastony:
*Issue #, if available:*
If two or more threads are handling the same ByteBuffer,
the ToDafny conversion layer can throw a `java.nio.BufferUnderflowException`.

This because the conversion method modifies the ByteBuffer's `position`.
Thus, 
other threads working on the same reference
may end up reading no bytes,
as the position is updated while they copying bytes
to the final output array.

*Description of changes:*


ajewellamz (approved):
Looks good.

PR #626, opened 2024-10-08
fix(Python): Support orphaned shapes, update Config shape generation

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

Python:
* Supported "orphaned" shapes (shapes with no graph connection to the LocalService)
* Refactor LocalService Config shape generation from its own logic to "normal" shape generation
  * Before, Config shapes were "orphaned" and could not be discovered by normal shape generation. I wrote manual generation for Config shapes.
  * Now, Config shape generation is more closely tied to normal shape generation, which gets it more fully-featured shape generation (constraints, as/from_dict)
* Don't use forward references for boto3 client typehints

Check out this MPL PR for sample generated code: https://github.com/aws/aws-cryptographic-material-providers-library/pull/831


codegen/smithy-dafny-codegen-modules/smithy-python/codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/StructureGenerator.java
```diff
@@ -252,7 +252,7 @@ protected boolean isOptionalDefault(MemberShape member) {
             && (target.isDocumentShape() || target.isListShape() || target.isMapShape());
     }
 
-    private void writeClassDocs(boolean isError) {
+    protected void writeClassDocs(boolean isError) {
```
lucasmcdonald3, line 255: Unfortunately, this change means another

`make mvn_local_deploy_polymorph_python_dependencies`

on merging this from main, but I think this is better than duplicating the code.

robin-aws: I'm inclined to add `mvn_local_deploy_polymorph_dependencies` as a dependency of all polymorph targets, given it should be cheap and hopefully Gradle-cached anyway.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/utils/ModelUtils.java
```diff
@@ -712,4 +714,54 @@ public static Optional<String> getDocumentationOrJavadoc(Shape shape) {
...
+
+    // orphanedShapes = (all shapes in model) - (non-orphaned shapes)
+    Set<Shape> orphanedShapes = model.shapes().collect(Collectors.toSet());
+    orphanedShapes.removeAll(nonOrphanedShapes);
+
+    // Copy-paste Smithy-Core's shape ordering mechanism for topological ordering
+    // (Python needs topological ordering to write shapes in order; other languages might not matter)
+    List<Shape> orderedShapes = new ArrayList();
+
+    TopologicalIndex topologicalIndex = TopologicalIndex.of(model);
```
robin-aws, line 735: Nice that this is already available for you!

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/utils/ModelUtils.java
```diff
@@ -712,4 +714,54 @@ public static Optional<String> getDocumentationOrJavadoc(Shape shape) {
...
+    // Get orphaned shapes under the following conditions:
+    // 1. In the same namespace as the service. (Should only generate shapes in this service.)
+    // 2. Not a member shape. (Member shapes don't have their own shapes generated for them.)
+    for (Shape shape : topologicalIndex.getOrderedShapes()) {
+      if (
+        orphanedShapes.contains(shape) &&
+        shape
+          .getId()
+          .getNamespace()
+          .equals(serviceShape.getId().getNamespace()) &&
```
robin-aws:
```suggestion
        ModelUtils.isInServiceNamespace(shape, serviceShape) &&
```

robin-aws: And likewise below. Just reads a bit better.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/utils/ModelUtils.java
```diff
@@ -712,4 +714,54 @@ public static Optional<String> getDocumentationOrJavadoc(Shape shape) {
...
+        !shape.isMemberShape()
+      ) {
+        orderedShapes.add(shape);
+      }
+    }
+    for (Shape shape : topologicalIndex.getRecursiveShapes()) {
```
robin-aws, line 749: I wasn't 100% sure from the TopologicalIndex documentation, but is this guaranteed to be a superset of `getOrderedShapes()`?

Would be nice to have an explicit unit test just for `getTopologicallyOrderedOrphanedShapesForService`, especially since it can clarify what is or isn't a recursive shape.

robin-aws (changes requested):
Quite happy with the general approach to handle orphaned shapes. Cut https://github.com/smithy-lang/smithy-dafny/issues/627 for longer term follow up on the problem.

Are you dependent on an OrphanedShapes test model to get added to test these changes?

codegen/smithy-dafny-codegen-modules/smithy-python/codegen/smithy-python-codegen/src/main/java/software/amazon/smithy/python/codegen/StructureGenerator.java
lucasmcdonald3: Good idea. I'll include this

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/utils/ModelUtils.java
lucasmcdonald3: > I wasn't 100% sure from the TopologicalIndex documentation, but is this guaranteed to be a superset of getOrderedShapes()?

I suspect they're disjoint, otherwise some shapes would be generated multiple times.

> Would be nice to have an explicit unit test just for getTopologicallyOrderedOrphanedShapesForService, especially since it can clarify what is or isn't a recursive shape.

I think a unit test might be more effort than it's worth --

Primarily, I'm inclined to trust whatever Smithy-Core is doing here, but verify it against a TestModel/real projects. I built this PR against the MPL and DBESDK, it picked up orphaned shapes, and I can run the code. I don't think the test you're proposing gets us as much ROI.

For a unit test, we'd have to create a Smithy model as a POJO (sort of like [this](https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/codegen/smithy-dafny-codegen/src/test/java/software/amazon/polymorph/util/TestModel.java)), which seems more brittle than is worth doing.

Let me know your thoughts.

lucasmcdonald3:
> Are you dependent on an OrphanedShapes test model to get added to test these changes?

I don't think so, but let me know if you disagree.

I built this PR into the MPL and DBESDK, and both of those projects are still passing.
* MPL PR: https://github.com/aws/aws-cryptographic-material-providers-library/pull/831
* DBESDK PR doesn't exist. This project only exists on my local right now.

This (plus the MPL PR) is ideally the last PR I'll to land before I release the Python MPL, so I'm trying to push any non-essential work out until after that lands.
I'm carving out some time to handle these post-launch fast follows, and I added OrphanedShapes TestModel to that list

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/utils/ModelUtils.java
robin-aws: > I suspect they're disjoint, otherwise some shapes would be generated multiple times.

Good point, I think they must be.

> For a unit test, we'd have to create a Smithy model as a POJO (sort of like [this](https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/codegen/smithy-dafny-codegen/src/test/java/software/amazon/polymorph/util/TestModel.java)), which seems more brittle than is worth doing.

You can include a *.smithy file as a resource and read it in pretty easily instead: https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/codegen/smithy-dafny-codegen/src/test/java/software/amazon/polymorph/smithyjava/modeled/ModeledShapeValueTest.java#L77-L82

But I'll leave it to your judgement on whether you want to bother with that on this PR, I won't block on it.

lucasmcdonald3: That's pretty nifty -- I'll plan to do this when I do the OrphanedShapes TestModel, but not now.

SmithyDafnyMakefile.mk
```diff
@@ -269,6 +272,7 @@ transpile_dependencies_test:
 #   make polymorph_code_gen CODEGEN_CLI_ROOT=/path/to/smithy-dafny/codegen/smithy-dafny-codegen-cli
 # StandardLibrary is filtered out from dependent-model patsubst list;
 #   Its model is contained in $(LIBRARY_ROOT)/model, not $(LIBRARY_ROOT)/../StandardLibrary/Model.
+_polymorph: mvn_local_deploy_polymorph_dependencies
```
robin-aws, line 275: ❤️

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/extensions/DafnyPythonLocalServiceStructureGenerator.java
```diff
@@ -45,17 +46,122 @@ public DafnyPythonLocalServiceStructureGenerator(
...
+   * called out with comments saying "Component below is changed from Smithy-Python."
+   */
+  protected void writeLocalServiceConfigShapeInit() {
+    writer.openBlock(
+      "def __init__(",
+      "):",
+      () -> {
+        writer.write("self,");
+        if (!shape.members().isEmpty()) {
+          // Adding this star to the front prevents the use of positional arguments.
```
robin-aws, line 108: Was this lifted from smithy-python too then? It's a great idea but I wasn't sure if they'd acted on it yet.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/extensions/DafnyPythonLocalServiceStructureGenerator.java
```diff
@@ -203,7 +309,14 @@ protected void writePropertyForMember(
       );
     }
 
-    if (target.hasTrait(ReferenceTrait.class)) {
+    // Reference shapes require forward reference to avoid circular import,
+    // but references to AWS SDKs don't
```
robin-aws, line 313: Nit picking, but isn't it more accurately that shapes from dependencies are already declared?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/wrappedlocalservice/extensions/DafnyPythonWrappedLocalServiceClientCodegenPlugin.java
```diff
@@ -41,6 +43,23 @@ public DafnyPythonWrappedLocalServiceClientCodegenPlugin(
...
+   * @param serviceShape
+   * @return
+   */
+  public static Model transformModelForWrappedLocalService(
+    Model model,
+    ServiceShape serviceShape
+  ) {
+    Model transformedModel = model;
+    transformedModel =
+      addWrappedLocalServiceTrait(transformedModel, serviceShape);
```
robin-aws, line 58: off topic question: what does this transformation do exactly?

robin-aws (approved):
LGTM, some non-blocking questions

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/wrappedlocalservice/extensions/DafnyPythonWrappedLocalServiceClientCodegenPlugin.java
lucasmcdonald3: Check out https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/wrappedlocalservice/extensions/DafnyPythonWrappedLocalServiceClientCodegenPlugin.java#L52

In preprocessing, it replaces a LocalServiceTrait on the ServiceTrait with a WrappedLocalServiceTrait.
This results in some different codegen; wrapped LocalServices don't get all the codegen from LocalServices. (No models, etc.)

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/extensions/DafnyPythonLocalServiceStructureGenerator.java
lucasmcdonald3: Yes

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/extensions/DafnyPythonLocalServiceStructureGenerator.java
lucasmcdonald3: This is actually about circular dependencies within one project.
I'm going to write about this problem and solutions in a Smithy-Dafny Python dev guide, but an overview/instance of the problem is that `models.py` can't import `references.py` at the top-level because `references.py` depends on `models.py` at the top-level, but `models.py` still needs to import `references.py` to use its classes.
One solution is to defer the imports like I've done here.
I wrote a fair amount about this and drew some graphs, but I'm going to collate all that later.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/wrappedlocalservice/extensions/DafnyPythonWrappedLocalServiceClientCodegenPlugin.java
robin-aws: Ah thanks! I wasn't aware of that trait.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/extensions/DafnyPythonLocalServiceStructureGenerator.java
robin-aws: I will be thrilled to review a PR that targets `docs/python` in the future :)

Issue #275, opened 2023-06-26
Design support for @streaming trait (including hand-written TestModel)


robin-aws:
Depends on https://github.com/dafny-lang/libraries/issues/70 to first define the idiomatic representation of streams in Dafny.

robin-aws:
Renamed this to reflect the need for defining what the Dafny interfaces to implement will look like, and how they will be adapted to the various kinds of streaming input/output types each target language supports.

kessplas:
Support for the `@streaming` trait entails support for both data streams and event streams as described in the [Smithy docs](https://smithy.io/2.0/spec/streaming.html). 

An example of an operation requiring streaming support is `S3::PutObject`:

```
        "com.amazonaws.s3#PutObject": {
            "type": "operation",
            "input": {
                "target": "com.amazonaws.s3#PutObjectRequest"
            },
            "output": {
                "target": "com.amazonaws.s3#PutObjectOutput"
            },
...
        "com.amazonaws.s3#PutObjectRequest": {
            "type": "structure",
            "members": {
                "ACL": {
                    "target": "com.amazonaws.s3#ObjectCannedACL",
                    "traits": {
                        "smithy.api#documentation": "[...]",
                        "smithy.api#httpHeader": "x-amz-acl"
                    }
                },
                "Body": {
                    "target": "com.amazonaws.s3#StreamingBlob",
                    "traits": {
                        "smithy.api#default": "",
                        "smithy.api#documentation": "<p>Object data.</p>",
                        "smithy.api#httpPayload": {}
                    }
                },
...
        "com.amazonaws.s3#StreamingBlob": {
            "type": "blob",
            "traits": {
                "smithy.api#streaming": {}
            }
        },
```

An example of code generated from this model can be found withIn the [AWS SDK Go v2](https://github.com/aws/aws-sdk-go-v2/blob/7ed7b4fc6595a7e554796424f3470b37287dde5c/service/s3/api_op_PutObject.go#L201); the `Body` member of the `PutObjectInput` structure is an `io.Reader`. 

In order to support S3 clients in the AWS SDK, Smithy-Dafny must support at least data streams to be able to generate Dafny code that can interact with services such as S3 which have `@streaming` traits in their models. 

[...]

PR #632, opened 2024-10-11
fix: minor tweaks to support S3 in Python

kessplas:
*Issue #, if available:* n/a

*Description of changes:* Three small changes to allow the AWS SDK S3 module to work in Python: 

1. Blob shapes which are streams must be read when converting from Python to Dafny, otherwise the blob's value is dropped. This requires buffering streams wholesale for now, which is consistent with other runtimes and will be addressed in 
https://github.com/smithy-lang/smithy-dafny/issues/275.
2. Adds a few other symbols to mangle when present in enumValues. S3 model has a wider set of enumValues which Python cannot handle. 
3. Loosen the validation on the enum trait. The S3 model does not adhere to the stricter regex, so we have to loosen it. A better solution might separate the validation for AWS SDK models vs. custom ones.


lucasmcdonald3 (approved):
I ran this against MPL and there were no changes

PR #653, opened 2024-10-22
chore: Add calling AWS SDK from local test model

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*
- Add AWS SDK from local service test model
- Done with an assumption that kmsv2 and ddbv2 will be working (I know few things are in PR rn)
- Disable this test model in every language. Will enable it for go in next PR which will include codegen changes necessary for this.


TestModels/CallingAWSSDKFromLocalService/test/SimpleCallingAWSSDKFromLocalServiceImplTest.dfy
```diff
@@ -0,0 +1,92 @@
...
+
+  method TestCallDDBScan_Failure(client: ISimpleCallingAWSSDKFromLocalServiceClient)
+    requires client.ValidState()
+    modifies client.Modifies
+    ensures client.ValidState()
+  {
+    var ddbClient :- expect Dynamodb.DynamoDBClient();
+    var resFailure := client.CallDDBScan(SimpleCallingAWSSDKFromLocalService.Types.CallDDBScanInput(ddbClient := ddbClient, tableArn := TABLE_ARN_FAILURE_CASE));
+
+    expect resFailure.Failure?;
```
rishav-karanjit: We can also check for Opaque error over here but first we need to check if its DynamoDB or KMS error for Dafny verification to pass and for specifically Opaque error its little fragile in Go codegen. So, it doesn't have it but lmk if you think otherwise.

ShubhamChaturvedi7: Let's add a TODO to verify the error that get's thrown? I guess this is blocked on our codegen error conversion fix?

TestModels/CallingAWSSDKFromLocalService/test/SimpleCallingAWSSDKFromLocalServiceImplTest.dfy
```diff
@@ -0,0 +1,92 @@
...
+      Plaintext := [ 97, 115, 100, 102 ],
+      EncryptionContext := Wrappers.None,
+      GrantTokens := Wrappers.None,
+      EncryptionAlgorithm := Wrappers.None
+    );
+    var resFailure_NonExistent := client.CallKMSEncrypt(SimpleCallingAWSSDKFromLocalService.Types.CallKMSEncryptInput(kmsClient := kmsClient, keyId := NONEXISTENT_KEY_ID, plaintext := PLAIN_TEXT));
+    
+    expect resFailure_NonExistent.Failure?;
+    expect resFailure_NonExistent.error.ComAmazonawsKms?;
+    expect resFailure_NonExistent.error.ComAmazonawsKms.NotFoundException?;
```
ShubhamChaturvedi7: Ohh I see it's here. Why wasn't this possible in DDB case?

TestModels/CallingAWSSDKFromLocalService/test/WrappedSimpleCallingAWSSDKFromLocalServiceTest.dfy
```diff
@@ -0,0 +1,33 @@
...
+  import opened WrappedSimpleCallingAWSSDKFromLocalServiceService
+  import SimpleCallingAWSSDKFromLocalServiceImplTest
+  import opened Wrappers
+  import opened StandardLibrary.UInt
+  import opened SimpleCallingawssdkfromlocalserviceTypes
+
+  const TABLE_ARN_FAILURE_CASE := "arn:aws:dynamodb:us-west-2:370957321024:table/TestTableFailure"
+  const NONEXISTENT_KEY_ID := "arn:aws:kms:us-west-2:658956600833:key/b3537ef1-d8dc-4780-9f5a-55776cbb2f7g"
+  // The string "asdf" as bytes
+  const PLAIN_TEXT: Kms.Types.PlaintextType := [ 97, 115, 100, 102 ]
```
ShubhamChaturvedi7: I think we aren't using these, remove them if we are not.

codegen/smithy-dafny-codegen-test/src/test/java/software/amazon/polymorph/smithygo/GoTestModels.java
```diff
@@ -17,6 +17,7 @@ class GoTestModels extends TestModelTest {
 
   static {
     DISABLED_TESTS.add("AggregateReferences");
+    DISABLED_TESTS.add("CallingAWSSDKFromLocalService");
```
ShubhamChaturvedi7: Why do we have it disabled for Go?

rishav-karanjit: Just to make CI pass. I will make a PR next to fix codegen next and remove this disabled test

rishav-karanjit: Also, I was initially think to make it as a single PR but thought that single PR could be too long. So, I divided that PR into 2 parts. This is 1st part and it is already 16 files long.

TestModels/CallingAWSSDKFromLocalService/test/SimpleCallingAWSSDKFromLocalServiceImplTest.dfy
rishav-karanjit: I changed my mind and added a check for opaque error. 

So, this is how codegen would work in next PR (also in comment)
- Go delegates the error first to KMS
- If KMS is opaque, it is going to delegate to DDB. And return a DDB error (opaque or non opaque)
- If KMS is not opaque, it returns ComAmazonawsKms error.

So, there is no KMS opaque error.

rishav-karanjit: Also, DDB give opaque error when trying to access table not in the account for user with access to limited number of table because they can't say resource not found (because user does not have access to all table) and they give AccessDeniedException which is not DDB exception but an IAM exception.

TestModels/CallingAWSSDKFromLocalService/test/SimpleCallingAWSSDKFromLocalServiceImplTest.dfy
rishav-karanjit: I think my previous comment would address this comment. lmk if it does not.

TestModels/CallingAWSSDKFromLocalService/test/WrappedSimpleCallingAWSSDKFromLocalServiceTest.dfy
rishav-karanjit: Yeah. I missed removing this while doing clean. My bad.

ShubhamChaturvedi7 (approved):

ShubhamChaturvedi7:
This PR LGTM. Let's enable the tests before merge.

rishav-karanjit:
Merging this because commits after approval is a reviewed PR or merge from main.

PR #662, opened 2024-10-24
fix: Support Windows properly

### LINKED ISSUE #317: Code generation not correct on Windows
Follow-up from issues with initial versions of https://github.com/aws/aws-cryptographic-material-providers-library/pull/195, which tried to execute `make polymorph_dafny` on windows but failed because `dafny format` hit weird errors on the generated code: https://github.com/aws/aws-cryptographic-material-providers-library/actions/runs/7905859646/job/21579405710?pr=195#step:10:1767

This is likely caused by not emitting the correct line separators: there are tons of hard-coded "\n" newlines in the code base, but it doesn't look onerous to replace them with `System.lineSeparator()`. I suspect the other Smithy code generators don't have this issue and hence I'm calling this a bug, even if we never claimed to support Windows.

robin-aws:
*Issue #, if available:*
Resolves #317 (AFAICT - the logs from the linked run have expired. But the CI now exercises codegen on Windows, which gives me confidence it should work. We can open new issues if we discover new issues)

*Description of changes:*
Addresses a handful of portability issues, so that Windows is supported:
1. Remove `check_dafny_version.sh` and extract/check the Dafny version in Java code instead (and updated `DafnyVersion` to handle the `+build` part of the semver 2.0). `--dafny-version` is now an optional parameter, provided just in case you want to override it.
2. Avoid the false error about the `project.properties` path containing a `:` (which it usually does on Windows), despite it not being a templated path. Also renamed an overload to the more specific `evalTemplateResource` to make the distinction clearer.
3. Replace `shell pwd` in the makefile with the more portable `$(CURDIR)`.

Added building a target on the SimpleString test model on Windows as a sanity check. I attempted building all models on all platforms, but besides this being expensive, my last attempt saw windows jobs hang for some reason. Would like to revisit that but later, as this PR definitely moves us forward.

Big thanks to @MikaelMayer for helping to figure out some of these fixes. :)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
```diff
@@ -1235,9 +1241,17 @@ public CodegenEngine build() {
...
-        this.dafnyVersion
-      );
+      final DafnyVersion dafnyVersion = Optional
+        .ofNullable(this.dafnyVersion)
+        .orElseGet(CodegenEngine::getDafnyVersionFromDafny);
+      if (dafnyVersion.compareTo(MIN_DAFNY_VERSION) < 0) {
+        throw new IllegalStateException(
+          "A minimum Dafny version of " +
+          MIN_DAFNY_VERSION.unparse() +
+          " is required"
```
alex-chew: nit: It'd be nice to report which version we found.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/utils/IOUtils.java
```diff
@@ -69,10 +69,9 @@ public static void writeTokenTreesIntoDir(
   /**
    * Evaluate a simple template as a resource under "/templates/<templatePath>"
    * and then write it to "<rootPath>/<templateOutputPath>".
-   * The template output path is also evaluated as in {@link #safeEvalPathTemplate(String, Map)}
```
alex-chew: This comment change isn't accurate, as it concerns the template output path and not the template resource path.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/utils/IOUtils.java
```diff
@@ -133,6 +136,21 @@ public static String evalTemplate(
     return evalTemplate(template, context);
   }
 
+  /**
```
alex-chew: Am I missing something or is this unused? (And duplicates `safeEvalPathTemplate`?

alex-chew (changes requested):

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/utils/IOUtils.java
robin-aws: Good catch, part of an earlier refactoring attempt to fix the problem. Deleted.

alex-chew (approved):

PR #663, opened 2024-10-24
feat: enable ddbv2 support

ShubhamChaturvedi7:
*Issue #, if available:*

*Description of changes:*


TestModels/aws-sdks/ddbv2/codegen-patches/dotnet/dafny-4.8.0.patch
robin-aws, line 1: FYI you don't need this file, patch files apply to all dafny versions at or above the named version automatically

TestModels/aws-sdks/kmsv2/runtimes/java/build.gradle.kts
robin-aws, line 1: FYI if you add `POLYMORPH_OPTIONS=--generate project-files` to the Makefile the `java` and `net` project files will be generated for you, so you don't have to check these in.

You can do `--generate client-constructors` too for DDB, but that doesn't work for KMS since we've added the extra `RegionMatch` method.

TestModels/aws-sdks/ddbv2/Model/model.json
robin-aws, line 1: Make sure you document why we do this really well in the README

TestModels/dafny-dependencies/Model/aws.test.smithy
robin-aws, line 1: What's this for?

TestModels/aws-sdks/ddbv2/runtimes/net/.editorconfig
robin-aws, line 1: Intentional checkin?

robin-aws (changes requested):

TestModels/aws-sdks/ddbv2/codegen-patches/dotnet/dafny-4.8.0.patch
ShubhamChaturvedi7: Nice! Removed.

TestModels/dafny-dependencies/Model/aws.test.smithy
ShubhamChaturvedi7: It's a trait used by the ddbv2 model - I was thinking to remove it but realized @texastony had this in his commit so I might as well pull it. Brings us one more step closer to full model support.

There's also a chance that we might be able to pull this using maven or something.

TestModels/aws-sdks/ddbv2/runtimes/net/.editorconfig
ShubhamChaturvedi7: Yeah - was present in ddb v1. So checked in here since this is not in the list of ignored files.

TestModels/dafny-dependencies/Model/aws.test.smithy
robin-aws: Ah okay, the cleaner way to pull it in is to add the appropriate dependency here: https://github.com/smithy-lang/smithy-dafny/blob/main-1.x/codegen/smithy-dafny-codegen/build.gradle.kts#L23

Note one advantage of the Smithy build plugin over the polymorph CLI is the former lets the Smithy project define these dependencies.

TestModels/aws-sdks/ddbv2/runtimes/net/.editorconfig
texastony: `.editorconfig` were my early attempt at getting .NET formatting and to ignore a bunch of errors.

Because .NET is so noisy.

TestModels/dafny-dependencies/Model/aws.test.smithy
texastony: I tried to remove the tests from the model,
but could not get it to work,
so I checked it in.

If we refactor the MPL/DB-ESDK/ESDK to use the Smithy Build Plugin,
then we can do as you say.

Otherwise, we should probably just check in this test file.
At least in the MPL.

ShubhamChaturvedi7: I think maven dependency should work for MPL as well.

TestModels/aws-sdks/kmsv2/runtimes/java/build.gradle.kts
ShubhamChaturvedi7: `POLYMORPH_OPTIONS=--generate project-files --generate client-constructors`
both of them cause more problems than they solve. The project-files doesn't use the right dependencies so the code will never compile unless we use another patch. client-constructors  modifies the Index.dfy which breaks the sed magic.

TestModels/aws-sdks/ddbv2/Model/model.json
ShubhamChaturvedi7: Added line explaining why are we doing this.

TestModels/dafny-dependencies/Model/aws.test.smithy
robin-aws: I can live with this for now

robin-aws (approved):

TestModels/dafny-dependencies/Model/aws.test.smithy
texastony: @ShubhamChaturvedi7 Alas, it will not, as the MPL does not use the Smithy Build Plugin.

ShubhamChaturvedi7: I think the context for Robin's smithy-build-plugin and yours is different. I think Robin meant that you can configure per project dependency IF we were using smithy-build. But what I'm doing is adding it to the codegen project. So it should work regardless if it's CLI or smithy-build.

PR #666, opened 2024-10-24
chore(docs/tests): SQSExtended test model (example of polymorph library calling an AWS service)

robin-aws:
*Description of changes:*

Adds a test model intended as an example of how to build a thick client using polymorphing, in this case adding a very common client-side operation for AWS SQS. My intention was to make this a `smithy init` template, but as it was significant work to get this example working I'm locking it down in CI first, then will find a way to make a template from it while still testing it.

Made several fixes and improvements to get this working in Java and make progress in other languages:

* Added a `aws.polymorph#localService$configRequired` parameter, defaulting to false. Previously all local services had to provide a default factory method for config structures, but this isn't always reasonable. For `SQSExtended` you really have to provide an underlying SQS client, and there's no sensible default value for an allocated service client.
* Found that Dafny only requires a `...types.__default.java` file if the `{:extern}` module has any subset types declared. To move closer to being able to eventually GENERATE ALL THE THINGS, added a dummy subset type in the `polymorph_dafny` output so this is always required, and included this empty file when doing `--generate client-constructors`. This isn't really breaking since all substantial polymorph libraries have lots of constraints in their models and hence subset types declared.
* Generalized the `BinarySetAttributeValue` handling to use `SdkBytes` correctly to all lists of blobs.
* Added a `dafnyError` case for `ToNative` on an arbitrary top-level service error (currently being patched in manually in the MPL).
* Fixed more naming shenanigans on error types in Java (and deleted a patch that was then no longer necessary).
* Added more SQS permissions to the CI role (already applied to the stack).
* Deleted some checked-in files that aren't necessary any more.

Also added a Setup section to the main README, since there are extra steps now.


TestModels/SQSExtended/src/SQSExtendedImpl.dfy
```diff
@@ -0,0 +1,61 @@
...
+    var maybeMessagesResult := config.sqsClient.ReceiveMessage(input.receiveRequest);
+    var messagesResult :- maybeMessagesResult.MapFailure(e => ComAmazonawsSqs(e));
+    // TODO: Handle error better
+    expect messagesResult.Messages.Some?;
+    var messages :- expect messagesResult.Messages;
+
+    for i := 0 to |messages| {
+      var message := messages[i];
+      // TODO: Propogate error instead of terminating
+      var receiptHandle :- expect message.ReceiptHandle;
```
seebees, line 38: I think you should just be able to `MapFailure` here. The smithy-dafny error type should have a joining error that should just work.

non-blocking point :)

seebees (approved):
This is super great!
Might have been better as multiple PRs, but I'm not sure if that is worth the overhead :)

TestModels/SQSExtended/src/SQSExtendedImpl.dfy
robin-aws: Yup agreed. I'll address that when I follow up with adding more target languages.

PR #668, opened 2024-10-26
Golang/reviewed

ShubhamChaturvedi7:
*Issue #, if available:*

*Description of changes:*
Implements the support for Golang in smithy-dafny.

Below commits are the list of all the features and they all are reviewed.


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
robin-aws, line 1: My understanding is that `smithygo` is to some degree a fork of https://github.com/aws/smithy-go - Can we add a README that points that out and identifies the commit we forked from?

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
```diff
@@ -242,6 +245,14 @@ private static Options getCliOptionsForBuild() {
           .hasArg()
           .build()
       )
+      .addOption(
+        Option
+          .builder()
+          .longOpt("output-go")
+          .desc("<optional> output directory for generated Go files")
```
robin-aws, line 252: Why is this optional for Go (and Python as well it seems)?

robin-aws (commented):
Just a couple of high-level documentation questions

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
ShubhamChaturvedi7: Because it is an optional parameter in dafny. If you don't provide this, the code gets generated wherever the dafny file lives.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/CodegenEngine.java
ShubhamChaturvedi7: Added

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
robin-aws: Sure, that's true of all the target languages as well. Just wondering whether there's a reason only Python and Go make it optional.

robin-aws (approved):
image

codegen/smithy-dafny-codegen-cli/src/main/java/software/amazon/polymorph/CodegenCli.java
ShubhamChaturvedi7: Because Python and Go do it the right way? :) . But yeah the reason was to have a better control - you can provide one if you want but you can also skip it (for prototyping and stuff).

PR #672, opened 2024-10-29
feat: Adds shape visitors for local service

ShubhamChaturvedi7:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
@@ -1,0 +1,686 @@
+package software.amazon.polymorph.smithygo.localservice.shapevisitor;
+
+import static software.amazon.polymorph.smithygo.codegen.SymbolUtils.POINTABLE;
+import static software.amazon.polymorph.smithygo.utils.Constants.DAFNY_RUNTIME_GO_LIBRARY_MODULE;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import software.amazon.polymorph.smithygo.codegen.GenerationContext;
+import software.amazon.polymorph.smithygo.codegen.GoWriter;
+import software.amazon.polymorph.smithygo.codegen.SmithyGoDependency;
+import software.amazon.polymorph.smithygo.codegen.knowledge.GoPointableIndex;
+import software.amazon.polymorph.smithygo.localservice.nameresolver.DafnyNameResolver;
+import software.amazon.polymorph.smithygo.localservice.nameresolver.SmithyNameResolver;
+import software.amazon.polymorph.smithygo.utils.GoCodegenUtils;
+import software.amazon.polymorph.traits.DafnyUtf8BytesTrait;
+import software.amazon.polymorph.traits.ReferenceTrait;
+import software.amazon.smithy.codegen.core.CodegenException;
+import software.amazon.smithy.model.shapes.BlobShape;
+import software.amazon.smithy.model.shapes.BooleanShape;
+import software.amazon.smithy.model.shapes.DoubleShape;
+import software.amazon.smithy.model.shapes.IntegerShape;
+import software.amazon.smithy.model.shapes.ListShape;
+import software.amazon.smithy.model.shapes.LongShape;
+import software.amazon.smithy.model.shapes.MapShape;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.shapes.ShapeVisitor;
+import software.amazon.smithy.model.shapes.StringShape;
+import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.model.shapes.TimestampShape;
+import software.amazon.smithy.model.shapes.UnionShape;
+import software.amazon.smithy.model.traits.EnumTrait;
+import software.amazon.smithy.utils.StringUtils;
+
+// TODO: Remove anonymous function in each of the shape visitor and test if it will work
+// TODO: if check with %s to figure out if it's a pointer or not and remove duplicate code when shape is optional vs non optional
+public class DafnyToSmithyShapeVisitor extends ShapeVisitor.Default<String> {
+
+  private final GenerationContext context;
+  private final String dataSource;
+  private final GoWriter writer;
+  private final boolean isConfigShape;
+  private final boolean isOptional;
+  //TODO: Ideally this shouldn't be static but with current design we need to access this across instances.
+  private static final Map<MemberShape, String> memberShapeConversionFuncMap =
+    new HashMap<>();
+
+  public DafnyToSmithyShapeVisitor(
+    final GenerationContext context,
+    final String dataSource,
+    final GoWriter writer,
+    final boolean isConfigShape
+  ) {
+    this(context, dataSource, writer, isConfigShape, false);
+  }
+
+  public DafnyToSmithyShapeVisitor(
+    final GenerationContext context,
+    final String dataSource,
+    final GoWriter writer,
+    final boolean isConfigShape,
+    final boolean isOptional
+  ) {
+    this.context = context;
+    this.dataSource = dataSource;
+    this.writer = writer;
+    this.isConfigShape = isConfigShape;
+    this.isOptional = isOptional;
+  }
+
+  public static Set<MemberShape> getAllShapesRequiringConversionFunc() {
+    return memberShapeConversionFuncMap.keySet();
+  }
+
+  public static void putShapesWithConversionFunc(
+    final MemberShape shape,
+    final String conversionFunc
+  ) {
+    memberShapeConversionFuncMap.put(shape, conversionFunc);
+  }
+
+  public static String getConversionFunc(final MemberShape shape) {
+    return memberShapeConversionFuncMap.get(shape);
+  }
+
+  protected String referenceStructureShape(final StructureShape shape) {
+    final ReferenceTrait referenceTrait = shape.expectTrait(
+      ReferenceTrait.class
+    );
+    final Shape resourceOrService = context
+      .model()
+      .expectShape(referenceTrait.getReferentId());
+    var namespace = "";
+    if (resourceOrService.asResourceShape().isPresent()) {
+      final var resourceShape = resourceOrService.asResourceShape().get();
+      if (
+        !resourceOrService
+          .toShapeId()
+          .getNamespace()
+          .equals(context.settings().getService().getNamespace())
+      ) {
+        writer.addImportFromModule(
+          SmithyNameResolver.getGoModuleNameForSmithyNamespace(
+            resourceOrService.toShapeId().getNamespace()
+          ),
+          SmithyNameResolver.shapeNamespace(resourceShape)
+        );
+        namespace =
+          SmithyNameResolver.shapeNamespace(resourceOrService).concat(".");
+      }
+      if (!this.isOptional) {
+        return "%s_FromDafny(%s)".formatted(
+            namespace.concat(resourceShape.toShapeId().getName()),
+            dataSource
+          );
+      }
+      return """
+      func () %s.I%s {
+          if %s == nil {
+              return nil;
+          }
+          return %s
+      }()""".formatted(
+          SmithyNameResolver.smithyTypesNamespace(resourceShape),
+          resourceShape.getId().getName(),
+          dataSource,
+          "%s_FromDafny(%s.(%s.I%s))".formatted(
+              namespace.concat(resourceShape.toShapeId().getName()),
+              dataSource,
+              DafnyNameResolver.dafnyTypesNamespace(resourceShape),
+              resourceShape.getId().getName()
+            )
+        );
+    } else {
+      final var serviceShape = resourceOrService.asServiceShape().get();
+      if (
+        !resourceOrService
+          .toShapeId()
+          .getNamespace()
+          .equals(context.settings().getService().getNamespace())
+      ) {
+        writer.addImportFromModule(
+          SmithyNameResolver.getGoModuleNameForSmithyNamespace(
+            resourceOrService.toShapeId().getNamespace()
+          ),
+          SmithyNameResolver.shapeNamespace(serviceShape)
+        );
+        namespace =
+          SmithyNameResolver.shapeNamespace(resourceOrService).concat(".");
+      }
+      if (!this.isOptional) {
+        return "return %1$s{%2$s}".formatted(
+            namespace.concat(
+              context.symbolProvider().toSymbol(serviceShape).getName()
+            ),
+            dataSource
+          );
+      }
+      return """
+      return func () *%s {
+          if %s == nil {
+              return nil;
+          }
+          return &%s{%s.(*%s)}
+      }()""".formatted(
+          namespace.concat(
+            context.symbolProvider().toSymbol(serviceShape).getName()
+          ),
+          dataSource,
+          namespace.concat(
+            context.symbolProvider().toSymbol(serviceShape).getName()
+          ),
+          dataSource,
+          DafnyNameResolver.getDafnyClient(serviceShape.toShapeId().getName())
+        );
+    }
+  }
+
+  @Override
+  protected String getDefault(final Shape shape) {
+    throw new CodegenException(
+      String.format(
+        "Unsupported conversion of %s to %s using the %s protocol",
+        shape,
+        shape.getType(),
+        context.protocolGenerator().getName()
+      )
+    );
+  }
+
+  @Override
+  public String blobShape(final BlobShape shape) {
+    writer.addImportFromModule(DAFNY_RUNTIME_GO_LIBRARY_MODULE, "dafny");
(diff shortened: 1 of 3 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
```diff
@@ -0,0 +1,686 @@
...
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.shapes.ShapeVisitor;
+import software.amazon.smithy.model.shapes.StringShape;
+import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.model.shapes.TimestampShape;
+import software.amazon.smithy.model.shapes.UnionShape;
+import software.amazon.smithy.model.traits.EnumTrait;
+import software.amazon.smithy.utils.StringUtils;
+
+// TODO: Remove anonymous function in each of the shape visitor and test if it will work
```
lucasmcdonald3, line 40: Do we need to do this now or track this? I don't think so, this seems less important, but just checking

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
```diff
@@ -0,0 +1,686 @@
...
+// TODO: if check with %s to figure out if it's a pointer or not and remove duplicate code when shape is optional vs non optional
```
lucasmcdonald3, line 41: Should this be a follow-up item? I don't fully understand this one

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
```diff
@@ -0,0 +1,686 @@
...
+    final String conversionFunc
+  ) {
+    memberShapeConversionFuncMap.put(shape, conversionFunc);
+  }
+
+  public static String getConversionFunc(final MemberShape shape) {
+    return memberShapeConversionFuncMap.get(shape);
+  }
+
+  protected String referenceStructureShape(final StructureShape shape) {
```
lucasmcdonald3, line 115: (nit) This is ~100 lines long -- maybe a

```
if serviceShape:
   serviceShape(resourceOrService.asServiceShape)
elif resourceShape:
   resourceShape(resourceOrService.asServiceShape)
else:
    throw ValueError("Unsupported reference shape: " + shape)
```
?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
```diff
@@ -0,0 +1,686 @@
...
+    final StringBuilder builder = new StringBuilder();
+    final MemberShape memberShape = shape.getMember();
+    final Shape targetShape = context
+      .model()
+      .expectShape(memberShape.getTarget());
+    final var symbol = context.symbolProvider().toSymbol(shape);
+    final Boolean assertionRequired = targetShape.isStructureShape();
+    builder.append(
+      """
+                           var fieldValue %s
```
lucasmcdonald3: Is this formatting off or this this the expected indentation?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
```diff
@@ -0,0 +1,686 @@
...
+    return """
+        %s
+        %s
+        return union
+    """.formatted(functionInit, eachMemberInUnion);
+  }
+
+  @Override
+  public String timestampShape(final TimestampShape shape) {
+    // TODO: Figure out timestamp types when working on timestampShape
```
lucasmcdonald3, line 752: I would prefer to raise an exception at codegen time if this is encountered -- safer to make codegen fail on unhandled shapes

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
```diff
@@ -0,0 +1,686 @@
...
+            .concat(shape.getId().getName())
+        )
+    );
+    for (final var memberShapeEntry : shape.getAllMembers().entrySet()) {
+      final var memberName = memberShapeEntry.getKey();
+      final var memberShape = memberShapeEntry.getValue();
+      final var targetShape = context
+        .model()
+        .expectShape(memberShape.getTarget());
+      //TODO: Is it ever possible for structure to be nil?
```
lucasmcdonald3, line 299: One approach to this lets you say the answer is "no":
If the structure is required, it will never be nil
If it is optional, you can check if it is nil before going into the structure converter:

```
nativeOutput = (input.isNotNil ? structureConverter(input) : nil)
```

and only go into the structure converter if it is not nil

But you might need to make a code change to get the optional checking in

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
```diff
@@ -0,0 +1,686 @@
...
+                context.symbolProvider().toSymbol(shape)
+              )
+            )
+            .concat(")");
+      }
+      final boolean assertionRequired =
+        memberShape.isOptional() &&
+        targetShape.isStructureShape() &&
+        !targetShape.hasTrait(ReferenceTrait.class);
+      final var derivedDataSource =
```
lucasmcdonald3, line 315: In general I'm struggling to understand what the generated code looks like in context, but this is probably the toughest one for me to grasp.

I think the easiest thing to do is link to some generated code? Do you have an MPL branch I can look at?

I don't know if this works with your timeline, but it would also help if you could leave comments above complicated code blocks with an inline example of what the generated code looks like, and maybe leave some more guiding comments?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
ShubhamChaturvedi7: Nope - this is just for our tracking and GoodToHave.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
ShubhamChaturvedi7: It is more nuanced than that. A member shape can be Optional but still not be a pointer. Nil check can only operate on pointer types and not on value types. A structure can be value type or pointer type in Go SDK depending on some set of rules - and that set of rules is not very well defined for Dafny vs SDK. Hence the TODO. We added this so that if something fails tomorrow for Structure being nil - we know where to look.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/SmithyToDafnyShapeVisitor.java
```diff
@@ -0,0 +1,724 @@
...
+          }
+      }()""";
+    return """
+    %s
+    %s
+    %s""".formatted(functionInit, eachMemberInUnion, defaultCase);
+  }
+
+  @Override
+  public String timestampShape(final TimestampShape shape) {
```
lucasmcdonald3, line 767: Ditto as in the reverse conversion -- raise exception at codegen time here?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/ShapeVisitorHelper.java
```diff
@@ -0,0 +1,151 @@
...
+public class ShapeVisitorHelper {
+
+  private static final Map<MemberShape, Boolean> optionalShapesToDafny =
+    new HashMap<>();
+
+  public static boolean isToDafnyShapeOptional(final MemberShape shape) {
+    return optionalShapesToDafny.get(shape);
+  }
+
+  public static String toNativeShapeVisitorWriter(
```
lucasmcdonald3, line 40: I think the methods in this file also need some Javadocs to help out readers.
I know what's going on here since I wrote something similar in Python, but it's probably not clear to anyone outside of you and me (and maybe the Rust folks?).
Plus, this isn't a standard thing in Smithy, so folks coming from Smithy won't know what this is.

Where is this generated? (Just the filename(s) should be fine)
What is it generating and why? (Conversion methods for shapes with members, this solves recursive shapes)

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/SmithyToDafnyShapeVisitor.java
```diff
@@ -0,0 +1,724 @@
...
+              if %s == nil {
+              return Wrappers.Companion_Option_.Create_None_()
+              }
+              return Wrappers.Companion_Option_.Create_Some_(%s)
+          }()""";
+        return goCodeBlock.formatted(dataSource, dataSource);
+      }
+    }
+
+    throw new UnsupportedOperationException(
```
lucasmcdonald3, line 179: Yeah, like this. Do this on the reverse function

lucasmcdonald3 (commented):
The more I digest this the more I like it.

If you could do a once-over to add some single-line comments inside some of the longer functions to help explain what's going on, that would help others understand it more quickly.

I would really like to see some generated code in context before I approve this but it seems reasonable.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
ShubhamChaturvedi7: For 36 - There are cases where we create something like:
```
return func() f1 {
return func() f2 {
// Implementation
}()
}()
```

This was because before handling the recursive shapes, we had anonymous function because all type conversion logic was inline, but after adding the support for recursive shapes - we continued using the anonymous function even though now the actual shapes are within their own functions.

For 37: There are bunch of places where we could reduce some of the repeated code-blocks (in Java / codegen) but pushed to future because it requires refactoring and might break the Go code.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
ShubhamChaturvedi7: Yeah it's hard to get a feel for generated code when there's so many template params to be filled in. But adding generated code might pollute the codegen and also end up becoming stale.

Here's a link the MPL that might give you some context: https://github.com/aws/aws-cryptographic-material-providers-library/blob/Golang/dev/AwsCryptographyPrimitives/runtimes/go/ImplementationFromDafny-go/awscryptographyprimitivessmithygenerated/to_native.go#L67

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
ShubhamChaturvedi7: The Go formatter takes care of this, but I've fixed it here as well.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
ShubhamChaturvedi7: AFAIK throwing exception fails the codegen on DDBv2 (new model specifically for Go) shapes, hence the `nil` part.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/SmithyToDafnyShapeVisitor.java
ShubhamChaturvedi7: Same issue here with the exception.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/ShapeVisitorHelper.java
ShubhamChaturvedi7: Added a class level javadoc explaining this is a delegator and helps avoid infinite recursion while generating typeconversion methods.

Also added relevant method javadocs.

ShubhamChaturvedi7:
> The more I digest this the more I like it.
> 
> If you could do a once-over to add some single-line comments inside some of the longer functions to help explain what's going on, that would help others understand it more quickly.
> 
> I would really like to see some generated code in context before I approve this but it seems reasonable.

Added some comments, renamed some variables to make it self explanatory, and did some better formatting to ease readability - our java formatter is really aggressive and breaks line more than what is generally done in Java. This hinders readability and makes method longer than what it actually should be.

Avoided adding lots of comments since it again gets stale and pollutes code (I favor self explanatory code than comments). Let me know if you think we can do few things better.

Here's an example of the generated code: https://github.com/aws/aws-cryptographic-material-providers-library/blob/Golang/dev/AwsCryptographyPrimitives/runtimes/go/ImplementationFromDafny-go/awscryptographyprimitivessmithygenerated/to_native.go#L67

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
lucasmcdonald3: Is `DafnyToSmithyShapeVisitor` used for both LocalService and AWS SDK shapes?

(Python only implements TimestampShape for AWS SDKs. It raises an exception at codegen time for LocalServices.)

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
lucasmcdonald3: Thanks! This is helpful

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/SmithyToDafnyShapeVisitor.java
lucasmcdonald3: Do we need Timestamps in LocalService?

lucasmcdonald3 (approved):
TimestampShape exceptions will be in a new PR

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
ShubhamChaturvedi7: Ahh I should be able to do that. I'll fix it up in a follow up PR since it's going to be again about ShapeVisitors.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/SmithyToDafnyShapeVisitor.java
ShubhamChaturvedi7: Responded above, will follow up in the next PR.

PR #673, opened 2024-10-30
chore: use --rust-module-name

ajewellamz:
*Issue #, if available:*

*Description of changes:*

Transpile Rust code with --rust-module-name, meaning that then generated implementation_from_dafny.rs file should not need to be edited; rather, a hand written lib.rs file is used instead.

The post-transpile patching steps were removed.

Most of the files changed are to bring each TestModel to the new scheme, i.e. we remove implementation_from_dafny.rs from the cargo.toml file and create a lib.rs file.

The version of Dafny used for testing was also bumped up to the current HEAD.


robin-aws (approved):
Great improvement, and happy the polymorph codegen didn't need to change because of the `use` declarations.

This does make the `patch-after-transpile` Java code dead code, but in this case I'd rather not delete it because I'd like to follow up with refactoring it to generate the `lib.rs` file instead in the near future.

PR #688, opened 2024-11-06
chore!: Python and Go codegen runs in `polymorph_code_gen` target

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:* Python and Go run as part of `polymorph_code_gen`.

**Breaking change:** The `dependency-library-name` CLI option has been removed. You must specify either/both of:
* `python-dependency-module-name`
* `go-dependency-module-name`

I have some downstream PRs open in Crypto Tools repos that use Python and/or Go codegen:
* https://github.com/aws/aws-cryptographic-material-providers-library/pull/1292
* https://github.com/aws/aws-encryption-sdk/pull/752


texastony:
@lucasmcdonald3 If you have the time, it would be great to also add Python to the `clean` recipe.  
For example, right now, when ever I need to re-polymorph the MPL, I run:
```
gmake clean;
rm -rf runtimes/python/src/aws_cryptographic_material_providers/smithygenerated
```

It should be straightforward to add the `rm -rf` of the Python Smithy generated to the Makefile.

lucasmcdonald3:
Putting this PR on ice until Go gets merged.

This would create hiccups for Go that, ideally, we'd avoid for now.

ShubhamChaturvedi7 (commented):
Changes look good, but let's wait till we have Go fully merged and has CI, which should be this week.

lucasmcdonald3:
Keeping this unmerged until Go ESDK is released.
The breaking change to the Makefile variables would cause a hiccup for Go, and we'd prefer to keep that moving at full speed.

PR #689, opened 2024-11-06
fix: Golang/timestamp

ShubhamChaturvedi7:
*Issue #, if available:*

*Description of changes:*
1. Aws sdk timestamp fix
2. Required timestamp in model
3. Formatting
4. Java and Rust CI would fail due to Positional (known Issue)

aws-sdk snippet:
```
func com_amazonaws_dynamodb_RestoreSummary_RestoreDateTime_FromDafny(input interface{}) *time.Time {
	return func() *time.Time {
		var s string

		for i := dafny.Iterate(input.(dafny.Sequence)); ; {
			val, ok := i()
			if !ok {
				break
			} else {
				s = s + string(val.(dafny.Char))
			}
		}
		if len(s) == 0 {
			panic("timestamp string is empty")
		} else {
			t, err := time.Parse("2006-01-02T15:04:05.999999Z", s)
			if err != nil {
				panic(err)
			}
			return &t
		}
	}()
}
```


TestModels/SimpleTypes/SimpleTimestamp/test/SimpleTimestampImplTest.dfy
```diff
@@ -4,42 +4,54 @@ include "../src/Index.dfy"
...
+  {
+    var dafnyTimestamp := "2024-06-11T12:34:56Z";
+    var ret :- expect client.GetTimestamp(SimpleTimestamp.Types.GetTimestampInput(value := Some(dafnyTimestamp)));
+    expect ret.value.Some?;
+    var retTimestamp := ret.value.value;
+
+    // It's okay if milliseconds are serialized, so only check for prefix
+    expect |retTimestamp| > 0;
+    expect retTimestamp[|retTimestamp| - 1] == 'Z';
+    expect dafnyTimestamp[|dafnyTimestamp| - 1] <= retTimestamp[|retTimestamp| - 1];
```
lucasmcdonald3, line 54: I don't think I get this -- what is this asserting?

TestModels/SimpleTypes/SimpleTimestamp/test/SimpleTimestampImplTest.dfy
```diff
@@ -4,42 +4,54 @@ include "../src/Index.dfy"
...
+    modifies client.Modifies
+    ensures client.ValidState()
+  {
+    var dafnyTimestamp := "2024-06-11T12:34:56.789Z";
+    var ret :- expect client.GetTimestampRequired(SimpleTimestamp.Types.GetTimestampRequiredInput(value := dafnyTimestamp));
+    expect ret.value == dafnyTimestamp;
+    print ret;
+  }
+
+  method TestGetTimestampNoMs(client: ISimpleTypesTimestampClient)
```
lucasmcdonald3:
```suggestion
  method TestGetTimestampNoMillis(client: ISimpleTypesTimestampClient)
```
Just clearer to me; I wasn't sure what a `NoM` was for a few moments

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/AwsSdkToDafnyShapeVisitor.java
```diff
@@ -654,15 +654,36 @@ public String unionShape(final UnionShape shape) {
...
+    if (isPointerType) {
+      nilCheck =
+        "if %s == nil {return %s}".formatted(dataSource, nilWrapIfRequired);
+    }
+
+    var conversionCode =
+      """
+      func () %s {
+        %s
+        formattedTime := %s.Format(\"2006-01-02T15:04:05.999999Z\")
```
lucasmcdonald3, line 677: I didn't realize we use this in the DescribeTable API.
Could you check what precision is reported by an API call to AWS SDK?

Given that we actually do use this in our code, I have some concern about precision. If AWS SDK reports millis and we convert it to "millis with microsecond precision", that's technically wrong.
(I don't think this is blocking, but I'd appreciate it if you could raise this to the Smithy-Dafny channel as part of the "we should figure out what's going on with Timestamps" story in case anyone thinks this possible imprecision is blocking.)

ShubhamChaturvedi7: There's no precision change here. Go doesn't increase the precision to microseconds (doesn't append extra `000`). The format is to bound it to microseconds.

TestModels/SimpleTypes/SimpleTimestamp/test/SimpleTimestampImplTest.dfy
ShubhamChaturvedi7: +1, just not sure if this will break anything. I'll accept and check if the CI passes.

TestModels/SimpleTypes/SimpleTimestamp/test/SimpleTimestampImplTest.dfy
ShubhamChaturvedi7: This diff came in due to `dafny format` I ran. AFAICT I think this is trying to compare two things. First that last character is `Z` and then that the `dafnyTimestamp` is prefix of `retTimestamp` - for cases where we end up adding extra `000` . But I think line 54 might be wrong and should be `expect dafnyTimestamp[..|dafnyTimestamp| - 1] <= retTimestamp[..|retTimestamp| - 1];`

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/AwsSdkToDafnyShapeVisitor.java
ShubhamChaturvedi7: Here's an output after conversion: `WrappedSimpleTypesTimestampTest.GetString:  {2024-06-11 12:34:56.789 +0000 UTC}
`

TestModels/SimpleTypes/SimpleTimestamp/test/SimpleTimestampImplTest.dfy
lucasmcdonald3: Ah! Makes sense

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/AwsSdkToDafnyShapeVisitor.java
lucasmcdonald3: Sweet -- I was very confused for a few moments; I thought that timestamp was from today and it looked like the day and month were swapped 😄

lucasmcdonald3 (approved):

ShubhamChaturvedi7:
Raised a PR for the fix `expect dafnyTimestamp[..|dafnyTimestamp| - 1] <= retTimestamp[..|retTimestamp| - 1];` https://github.com/smithy-lang/smithy-dafny/pull/691/files

PR #709, opened 2024-11-11
feat(Python)!: Python codegen raises error if formatters aren't installed

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

**BREAKING CHANGE:** You must install `docformatter` and `black` to run Python codegen. Smithy-Dafny Python codegen will now raise an error when these formatters are not found. This was previously treated as a warning. This change may disrupt existing workflows that generated Python code without these formatters installed.


rishav-karanjit (approved):
LGTM

PR #714, opened 2024-11-12
fix: change function name for positional trait

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*
The conversion function (to native and to dafny) in local service of member shape with positional trait in its container has to be exposed. PR to fix it. 


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/utils/Constants.java
```diff
@@ -9,7 +13,33 @@ public class Constants {
...
+    final Model model
+  ) {
+    StringBuilder funcName = new StringBuilder(
+      funcNameGenerator(memberShape, suffix)
+    );
+    Shape containerShape = model.expectShape(memberShape.getContainer());
+    // membershape inside a container shape with positional trait has to be exposed.
+    if (containerShape.hasTrait(PositionalTrait.class)) {
+      funcName.setCharAt(0, Character.toUpperCase(funcName.charAt(0)));
+    }
```
ShubhamChaturvedi7: finalize and use `StringUtils.capitalize()`

ShubhamChaturvedi7 (approved):

PR #726, opened 2024-11-18
fix: Assert datasource in to native conversion of dafny client

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*
Assertion was missing in to native conversion of Dafny client when it is non optional. This PR adds it. 

Fixes following MPL error:
```
../../../../../aws-cryptographic-material-providers-library/AwsCryptographicMaterialProviders/runtimes/go/ImplementationFromDafny-go/awscryptographymaterialproviderssmithygenerated/to_native.go:2548:56: 
cannot use input (variable of type interface{}) as *KeyStore.KeyStoreClient value in struct literal: need type assertion
```

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
@@ -171,11 +171,12 @@ public class DafnyToSmithyShapeVisitor extends ShapeVisitor.Default<String> {
           );
       }
       if (!this.isOptional) {
-        return "return &%s{%s}".formatted(
+        return "return &%s{%s.(*%s)}".formatted(
             namespace.concat(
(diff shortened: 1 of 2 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
```diff
@@ -171,11 +171,12 @@ protected String referenceStructureShape(final StructureShape shape) {
           );
       }
       if (!this.isOptional) {
-        return "return &%s{%s}".formatted(
+        return "return &%s{%s.(*%s)}".formatted(
```
ShubhamChaturvedi7: I thought this would get fixed by using concrete types over `interface{}` for Positional. Is it not?

ShubhamChaturvedi7 (approved):
Discussed offline - it's due to TypeConversion design and not Positional.

rishav-karanjit:
Merging this as after the approval there is no change in the diff of the PR (just merging the latest changes in `Golang/dev`

PR #729, opened 2024-11-21
Fix: fixes in Go codegen for bug found in test vectors

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*
- In validation generator, we are calling `<dataSource>.Validate` without a nil check. In some cases, when <dataSource> is optional, it can be nil as well and `nil.Validate` will panic. So, fixed it with a if check. We don't need this check for list, map or union because when we visit list, map or union we are going to for loop or switch case which will eliminate the risk of being nil.
- In validation generator class, we had a check on `GoPointableIndex.of(model).isPointable(memberShape)` before we do a nil check for shapes with required trait. However, list, map and union can be nil even if they are not a pointer. So, I removed the isPointable check and replaced with list, map and union check. I believe aws sdk shapes can have a pointer even if it is required which is not in our control. So, there is a check to check if the shape is from aws sdk and if it is optional.
- We were using `DafnyNameResolver.getDafnyClient` on api client, type conversion layer, and other places. For all the cases (especially with model with dependency), this will not be true and we need to use a interface instead of a concrete type. So, changed this to Interface (`getDafnyInterfaceClient`). Since, this is a interface a pointer is also not required. 
- In a type conversion layer, we were doing `.UnwrapOr(nil).(type-assertion)`. This will fail when `UnwrapOr(nil)` will return a nil. Also a type assertion is not required here because in downstream function accept a `interface{}`.
- Do a nil check on list and union if it is optional (Note list and union are always non pointer type).


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/codegen/ValidationGenerator.java
```diff
@@ -398,14 +420,26 @@ private StringBuilder addRequiredCheck(
     ) {
```
ShubhamChaturvedi7: Why do we care about targetShape?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/codegen/ValidationGenerator.java
```diff
@@ -398,14 +420,26 @@ private StringBuilder addRequiredCheck(
...
-      );
+    final String nilCheck =
+      """
+      if ( %s == nil ) {
+          return fmt.Errorf(\"%s is required but has a nil value.\")
+      }
+      """;
+    if (
+      SmithyNameResolver.isShapeFromAWSSDK(targetShape) &&
+      memberShape.isOptional()
```
ShubhamChaturvedi7: You can use `AwsPointableIndex` but I'm also OK with this if that causes more problem. My concern is in a shape is Optional in aws-sdk but isn't pointable.

ShubhamChaturvedi7: On a second though line 418 is opposite of 431 right? So this will never be true?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
```diff
@@ -318,6 +328,18 @@ public String structureShape(final StructureShape shape) {
     final String referenceType = (this.isOptional) ? "&" : "";
 
     final var typeConversionMethodBuilder = new StringBuilder();
+    if (this.isOptional) {
+      final String unAssertedDataSource = dataSource.startsWith("input.(")
+        ? "input"
+        : dataSource;
+      typeConversionMethodBuilder.append(
```
ShubhamChaturvedi7, line 335: You might be able to use regex and just check for `.(%s)` and then split at that. Makes it robust against changing the name `input`.

ShubhamChaturvedi7 (approved):
Made a small change.

ShubhamChaturvedi7 (approved):

PR #732, opened 2024-11-26
feat: Merge from Golang/dev

ShubhamChaturvedi7:
*Issue #, if available:*

*Description of changes:*
Pulls in bug fixes from Golang/dev

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
@@ -13,7 +13,9 @@ import software.amazon.polymorph.smithygo.localservice.nameresolver.DafnyNameRes
 import software.amazon.polymorph.smithygo.localservice.nameresolver.SmithyNameResolver;
 import software.amazon.polymorph.smithygo.utils.GoCodegenUtils;
 import software.amazon.polymorph.traits.DafnyUtf8BytesTrait;
+import software.amazon.polymorph.traits.PositionalTrait;
 import software.amazon.polymorph.traits.ReferenceTrait;
+import software.amazon.smithy.aws.traits.ServiceTrait;
 import software.amazon.smithy.codegen.core.CodegenException;
 import software.amazon.smithy.model.shapes.BlobShape;
 import software.amazon.smithy.model.shapes.BooleanShape;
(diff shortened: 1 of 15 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
```diff
@@ -153,12 +156,32 @@ private String referencedServiceShape(final Shape resourceOrService) {
...
+      """.formatted(
+          dataSource,
+          DafnyNameResolver.dafnyNamespace(
+            resourceOrService.expectTrait(ServiceTrait.class)
+          )
+        );
+    }
     if (!this.isOptional) {
-      return "return %1$s{%2$s}".formatted(
+      return "return &%s{%s.(*%s)}".formatted(
```
josecorella, line 179: i'm not too well versed in go, but what is the difference here between
`&%s`
`%s`
`*%s`
?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
```diff
@@ -474,46 +553,93 @@ public String booleanShape(final BooleanShape shape) {
...
+              shape,
+              context.symbolProvider().toSymbol(shape)
+            )
+          );
+      } else {
+        return """
+          return func () %s.%s {
+          var u %s.%s
+        	inputEnum := %s.(%s)
+        	index := -1;
```
josecorella, line 611: why does index start at -1? no 0?

josecorella (commented):
looks good, have two questions but after that we should be able to merge

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
ShubhamChaturvedi7: `&` is the reference operator and `*` is de-reference / pointer operator. `%s` is the java formatting placeholder which fills the value using the arguments supplied to the `formatted()` method.

In this context it creates something like `&Name{input.(*string)}` (just an example and not executable code) which means that assert that `input` is `pointer type to string` and create `structure Name` out of it and return the reference (address) of the created structure.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
ShubhamChaturvedi7: We assign `-1` so that while iterating if none of the values match (line 616) we don't end up returning first value in the enum (line 622), rather panic / throw exception.

josecorella: yup! as soon as i got up from my desk and walked around I realized why, nice

josecorella (approved):

PR #734, opened 2024-12-02
feat: smithy-go

ShubhamChaturvedi7:
*Issue #, if available:*

*Description of changes:*


codegen/smithy-dafny-codegen/build.gradle.kts
```diff
@@ -1,6 +1,8 @@
 // Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import java.util.regex.Pattern.compile
+
```
texastony: Suggestion: Remove unused import. 
Presumably this was for a since abandoned idea. 
```suggestion
```

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/codegen/ValidationGenerator.java
```diff
@@ -196,10 +200,28 @@ private void renderValidatorForEachShape(
         break;
       case STRUCTURE:
         if (!currentShape.hasTrait(ReferenceTrait.class)) {
+          final boolean needsWrapping =
+            memberShape.isOptional() &&
+            !dataSource.equals(LIST_ITEM) &&
+            !dataSource.equals(MAP_KEY) &&
+            !dataSource.equals(MAP_VALUE) &&
+            !dataSource.equals(UNION_DATASOURCE);
```
texastony: Suggestion: `needsWrapping` is confusing to me.
We have Native Wrappers, we often talk about wrapping errors, or wrapping clients.
It looks like this a check if something can be null.
I suggest calling it `maybeNull`.
(Assuming `nil` is Go's allegory for `null`.

```suggestion
          final boolean maybeNull =
            memberShape.isOptional() &&
            !dataSource.equals(LIST_ITEM) &&
            !dataSource.equals(MAP_KEY) &&
            !dataSource.equals(MAP_VALUE) &&
            !dataSource.equals(UNION_DATASOURCE);
```

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/codegen/ValidationGenerator.java
```diff
@@ -196,10 +200,28 @@ private void renderValidatorForEachShape(
...
           final var funcCall = dataSource.concat(".Validate()");
-          validationCode.append(
-            CHECK_AND_RETURN_ERROR.formatted(funcCall, funcCall)
+          final String checkForError = CHECK_AND_RETURN_ERROR.formatted(
+            funcCall,
+            funcCall
           );
+          if (needsWrapping) {
```
texastony:
```suggestion
          if (maybeNull) {
```

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/DafnyLocalServiceGenerator.java
```diff
@@ -152,9 +155,7 @@ func NewClient(clientConfig $L) (*$T, error) {
           operation,
           OperationShape.class
         );
-        final var inputShape = model.expectShape(
-          operationShape.getInputShape()
-        );
+        var inputShape = model.expectShape(operationShape.getInputShape());
```
texastony, line 158: I honestly did not know that Java had `var` keyword.
Apparently, I have only ever written for Java 1.8, and Java 10 introduced `var`.
Wild.
(I hate it. Strong typing for the win. But I am not blocking/suggesting changes for it.)

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/DafnyLocalServiceGenerator.java
```diff
@@ -303,8 +338,29 @@ func NewClient(clientConfig $L) (*$T, error) {
                     symbolProvider.toSymbol(outputShape)
                   )
                 );
+            returnError = "return nil,";
           }
-          returnError = "return nil,";
+        }
+        String validationCheck = "";
+        // inputShape is not a structure only when we have a positional trait
+        // TODO: Figure out how to validate constraint in a structure with positional trait
```
texastony, line 346: This is a general Smithy-Dafny gap... I think... 
I hope that we deprecate positional.

It goes against Smithy's philosophy and is PITA to implement.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/DafnyLocalServiceGenerator.java
```diff
@@ -303,8 +338,29 @@ func NewClient(clientConfig $L) (*$T, error) {
...
+        if (
+          !inputShape.hasTrait(UnitTypeTrait.class) &&
+          inputShape.isStructureShape()
+        ) {
+          validationCheck =
+            """
+                err := params.Validate()
+                if err != nil {
+                  opaqueErr := %s.OpaqueError{
```
texastony, line 355: Out of curiosity, is the behavior in Go to return an Opaque Error for invalid input?

In Java, we thrown an `IllegalArguementException`. 
I do not know what .NET or Python do.

You could throw the local service's default exception.
i.e: in the MPL, [`AwsCryptographicMaterialProvidersException`](https://github.com/aws/aws-cryptographic-material-providers-library/blob/89701fe7582f336ab7caddc25c256b3bf2aaf1c1/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/Model/material-provider.smithy#L83-L87).

That would be marginally better.
Because `Opaque` is intended for unknown errors.
Where as we clearly know that root cause here: bad input.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/DafnyLocalServiceGenerator.java
```diff
@@ -633,30 +740,29 @@ void generateReferencedResources(GenerationContext context) {
...
+                        symbolProvider
+                      );
                     w.write(
                       """
                       $L($L) ($L error)
                       """,
                       operationShape.getId().getName(),
                       input,
-                      output
+                      output.equals("") ? "" : output.concat(",")
```
texastony: Nit: add a comment. 
I _think_ this is for a void/Unit response.
Or make the ternary's switch be `hasTrait(UnitTypeTrait.class)`.
That would be easier to follow, but more typing/characters.

Either is better than making me scratch my head for a minute or two...

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/DafnyLocalServiceGenerator.java
```diff
@@ -763,27 +882,95 @@ void generateReferencedResources(GenerationContext context) {
...
+                      : "";
+                    returnResponse =
+                      """
+                      var native_response = %s(dafny_response.Dtor_value()%s)
+                      return %snative_response, nil
+                      """.formatted(
+                          fromDafnyConvMethodName,
+                          typeAssertion,
+                          // Currently we expect all the structure with positional shape to have required trait on its membershape.
+                          // TODO: If this changed in https://github.com/smithy-lang/smithy-dafny/issues/727, we need to fix it.
```
texastony, line 944: I think we have an internal SIM about this as well...

codegen/smithy-dafny-codegen/src/main/resources/META-INF/services/software.amazon.polymorph.smithygo.codegen.integration.GoIntegration
texastony, line 1: What is this?
Did you mean to check this in?

texastony (approved):
I have a few nits, but nothing blocking.

The Java looks good to me.
I have no idea about the quality of the `Go` it creates... 
that is beyond my ability to ascertain in a PR.

@ShubhamChaturvedi7 is there a Test Model for Positional?
Or for Unit?

ShubhamChaturvedi7:
> I have a few nits, but nothing blocking.
> 
> The Java looks good to me. I have no idea about the quality of the `Go` it creates... that is beyond my ability to ascertain in a PR.
> 
> @ShubhamChaturvedi7 is there a Test Model for Positional? Or for Unit?

There is: https://github.com/smithy-lang/smithy-dafny/tree/Golang/dev/TestModels/Positional . I'm copying over logical pieces of the code into different PRs so that I can distribute it to different people and not have a super long PR.

You can go through this if you want to get a sense of the generated Go code: https://github.com/aws/aws-cryptographic-material-providers-library/tree/Golang/preview/AwsCryptographicMaterialProviders/runtimes/go/ImplementationFromDafny-go

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/codegen/ValidationGenerator.java
ShubhamChaturvedi7: Agreed. Me and @rishav-karanjit had few instances of naming-and-renaming.

But I think `needsWrapping` here means we want to return a `Wrapper` and not naked type. But then I look at the code and it seems like it is actually just doing a null check. @rishav-karanjit ?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/DafnyLocalServiceGenerator.java
ShubhamChaturvedi7: I discovered `var` last year when I started this project. I was in the camp Strong typing but then realized more and more languages are dynamically typed (at least in syntactical sense). Plus this is not a prod code. (I still prefer production have the types just for readability/debugging)

codegen/smithy-dafny-codegen/src/main/resources/META-INF/services/software.amazon.polymorph.smithygo.codegen.integration.GoIntegration
ShubhamChaturvedi7: This was to allow smithy build plugin to run. I think we are re-writing it anyway and none of the other languages support this integration way so I can drop it.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/DafnyLocalServiceGenerator.java
ShubhamChaturvedi7: Prefer the GHI since external customers can also read it.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/DafnyLocalServiceGenerator.java
ShubhamChaturvedi7: It's a bit more complicated in Go.

Our customer facing API doesn't impose any checks on what errors we can return. But our wrapped shim does. And Dafny Go doesn't use generics so there's no way to know what error we have received in the shim.

There are bunch of work arounds and we could use the base exception but I think there's not much benefit to reap compared to the re-architecting that we might have to do.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/codegen/ValidationGenerator.java
rishav-karanjit: Yes. This does null check.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/DafnyLocalServiceGenerator.java
ShubhamChaturvedi7: Added a comment with isEmpty() check

PR #742, opened 2024-12-10
fix: escape reserved words in Rust

ajewellamz:
*Issue #, if available:*

*Description of changes:*


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustUtils.java
```diff
@@ -1,12 +1,75 @@
...
+    "use",
+    "virtual",
+    "where",
+    "while",
+    "yield"
+  );
+
+  public static String escapedName(String s) {
+    if (RESERVED_WORDS.contains(s)) {
+      return "r#" + s;
```
robin-aws, line 67: Out of curiosity, is there any downside to just using `r#` all the time?

robin-aws (approved):
Would be ideal to have an explicit unit test, but it's low-value enough I don't insist on it in this case.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustUtils.java
ajewellamz: It's just ugly.

PR #744, opened 2024-12-11
chore: OrphanedShapes TestModel, GHA uses ubuntu-22.04

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

OrphanedShapes TestModel + some missing Python conversions.

The bar I'm aiming for is "orphaned shapes are supported if an extern can convert an orphaned shape to/from Dafny".
This can't be tested via local services because passing a shape in a service's operations makes that shape no longer orphaned.
Externs are a reasonable way to test between native/Dafny code, and shapes passed via extern operations can be orphaned.


TestModels/OrphanedShapes/README.md
```diff
@@ -0,0 +1,57 @@
+# OrphanedShapes
+
+## Background
```
robin-aws, line 3: Most of this content should exist outside of a specific TestModel IMO. I'd suggest moving a good chunk into a separate file under `designs`.

I'd also suggest naming the complete list of use cases that leads us to deviate from Smithy-Core's approach, just to be clear about which are fundamental (like config shapes) and which are closer to bugs in our implementations.

TestModels/OrphanedShapes/README.md
```diff
@@ -0,0 +1,57 @@
...
+- Aggregate
+- Enum
+
+and these TestModels' prerequisite TestModels.
+
+## Coverage
+
+This TestModel tests some instances of orphaned shapes
+
+- LocalService Config shapes. (Config shapes are "orphaned", but are likely already handled as one-offs by any codegen that's this TestModels' prerequisites)
```
robin-aws, line 48: This is the main "legit" reason to deviate at least, and I'm hoping in the future we can model local service client creation as an operational as well somehow, since then it wouldn't be a special case.

TestModels/OrphanedShapes/README.md
```diff
@@ -0,0 +1,57 @@
...
+- Errors
```
robin-aws, line 49: Why are these orphaned? Is it because errors from dependency services are used but not directly connected?

TestModels/OrphanedShapes/README.md
```diff
@@ -0,0 +1,57 @@
...
+- Resources (with @aws.polymorph#reference trait) and their operations
```
robin-aws, line 50: These should still be listed as `resources` on the local service though, right?

TestModels/OrphanedShapes/README.md
```diff
@@ -0,0 +1,57 @@
...
+- Structures (and structures' members)
```
robin-aws, line 51: Which structures other than config structures?

TestModels/.gitignore
```diff
@@ -34,6 +34,23 @@
 # Polymorph Generated Python
 **/runtimes/python/src/**/smithygenerated/
 
+# Polymorph Generated Rust
```
robin-aws, line 37: Thanks, what a pain to maintain but nice to have

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/common/shapevisitor/conversionwriter/BaseConversionWriter.java
```diff
@@ -121,4 +129,34 @@ protected abstract void writeStructureShapeConverter(
   protected abstract void writeStringEnumShapeConverter(
     StringShape stringShapeWithEnumTrait
   );
+
+  /**
+   * Returns true if a conversion function should be written for the shape, false otherwise.
+   * Conversion functions are only written for "complex" shapes:
+   *  - StructureShapes ("complex" because StructureShapes can be recursive)
+   *    - except for non-AWS SDK StructureShapes with ErrorTrait; these aren't "complex"
```
robin-aws, line 137: This won't be true forever as we should lift the smithy-dafny restriction that errors can only have a single string "message" member

robin-aws (commented):
Partial review, looking to clarify the top-level need for the feature first before reviewing the rest

TestModels/OrphanedShapes/README.md
lucasmcdonald3: Errors will be orphaned if they are not attached to any operations

TestModels/OrphanedShapes/README.md
lucasmcdonald3: Maybe, but in many cases they aren't

lucasmcdonald3: (I also believe that Smithy-Core doesn't discover resources attached to the service -- it's only mixins, operations, and errors)

TestModels/OrphanedShapes/README.md
lucasmcdonald3: The only examples are unused structures in the MPL.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/common/shapevisitor/conversionwriter/BaseConversionWriter.java
lucasmcdonald3: Sure -- I think that's a "cross that bridge when we come to it" situation

TestModels/OrphanedShapes/README.md
lucasmcdonald3: I can do some content moving.

I don't like writing a complete list anywhere. It's a long list (I'd estimate 30+) and it will get stale quickly. This list has recently gotten longer due to Tony's mutations changes.

TestModels/OrphanedShapes/README.md
robin-aws: That wouldn't surprise me given no other code generators produce anything for resources yet. I would still argue Smithy-Core probably should traverse them even if no one else does anything for them yet, so I might try to submit that change to Smithy-Core down the road...

TestModels/OrphanedShapes/README.md
robin-aws: Right, but I assume we still need conversions for them for some reason, just trying to understand that better

TestModels/OrphanedShapes/README.md
robin-aws: 30+ individual reasons we need orphaned shapes? Yikes!

I guess what you're saying is: if I wanted to get to a place where local services didn't need generation for orphaned shapes, I'd be better off building that and testing it against all the existing libraries to identify gaps?

(Thanks for satisfying my curiousity on these question btw, I have no intention of blocking this PR as we clearly should be explicitly testing a feature we depend on, I just want to take the opportunity to understand the scope of the feature better, since you would have had to dive into it a certain depth anyway)

TestModels/OrphanedShapes/README.md
robin-aws: Which ones specifically? I don't yet appreciate the use case

TestModels/OrphanedShapes/README.md
lucasmcdonald3: NVM! Resources attached to services are discovered
https://github.com/smithy-lang/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/neighbor/NeighborVisitor.java#L89-L91

TestModels/OrphanedShapes/README.md
lucasmcdonald3: There's this one https://github.com/aws/aws-cryptographic-material-providers-library/blob/23650a9c6e1c897455e2d59d15b7d7a2c5a07cdd/AwsCryptographyPrimitives/Model/aes.smithy#L62-L69

There's some unions too https://github.com/aws/aws-cryptographic-material-providers-library/blob/23650a9c6e1c897455e2d59d15b7d7a2c5a07cdd/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/Model/key-agreement-scheme.smithy#L5-L28

Then there may also be descendents of Config shapes, if Config shapes have orphaned members, like in DBESDK https://github.com/aws/aws-database-encryption-sdk-dynamodb/blob/main/DynamoDbEncryption/dafny/DynamoDbEncryption/Model/DynamoDbEncryption.smithy#L145-L210

TestModels/OrphanedShapes/README.md
lucasmcdonald3: In my mind, it's because Java generates conversions for orphaned errors, so other languages should also generate conversions for parity.

TestModels/OrphanedShapes/README.md
lucasmcdonald3: Yes, I think so.

I think we're landing on the idea of the "ideal state", which would be using Smithy-Core's shape discovery logic plus a bounded, limited set of orphaned shapes (like local service config shapes).
And the path to get there may just be a breaking change to Smithy-Dafny (made by testing against complex libraries) combined with Smithy model changes to downstream libraries

TestModels/OrphanedShapes/runtimes/rust/src/extern_definitions.rs
```diff
@@ -0,0 +1,75 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::rc::Rc;
+use std::cell::UnsafeCell;
+use dafny_runtime::rcmut;
+use crate::r#_Wrappers_Compile::Result;
+
```
ajewellamz, line 8: If we add
`use crate::simple::orphaned::internaldafny::types`
 then we can replace `use crate::simple::orphaned::internaldafny::types` with `types` below and things will be easier to read.

lucasmcdonald3: I tried this out but it conflicts with the regular (not internaldafny) types file -- I may just leave this as-is unless there's a workaround

lucasmcdonald3: Workaround was to alias the import

TestModels/OrphanedShapes/runtimes/go/TestsFromDafny-go/ExternDefinitions/extern.go
```diff
@@ -0,0 +1,39 @@
...
+
+	Wrappers "github.com/dafny-lang/DafnyStandardLibGo/Wrappers"
+	OrphanedResource "github.com/smithy-lang/smithy-dafny/TestModels/OrphanedShapes/OrphanedResource"
+	SimpleOrphanedTypes "github.com/smithy-lang/smithy-dafny/TestModels/OrphanedShapes/SimpleOrphanedTypes"
+	simpleorphanedsmithygenerated "github.com/smithy-lang/smithy-dafny/TestModels/OrphanedShapes/simpleorphanedsmithygenerated"
+	simpleorphanedsmithygeneratedtypes "github.com/smithy-lang/smithy-dafny/TestModels/OrphanedShapes/simpleorphanedsmithygeneratedtypes"
+)
+
+var _ Wrappers.Dummy__
+
```
rishav-karanjit, line 14: Non blocking suggestion because Go is in TODO anyway:

This is because there is a bug in dafny transpiled code. The transpiled code will be like `m_ExternDefinitions.InitializeOrphanedStructure(...)` however it should be only `InitializeOrphanedStructure(...)`
```suggestion

// TODO: Remove this once Dafny bug is fixed.
// Dafny bug: https://t.corp.amazon.com/9a9028fd-2711-4843-b8f0-09965f7e61a7/communication#f03694be-7410-47f9-866d-e43093b018f9
var m_ExternDefinitions = CompanionStruct_Default___{}
```

TestModels/OrphanedShapes/runtimes/go/TestsFromDafny-go/ExternDefinitions/extern.go
```diff
@@ -0,0 +1,39 @@
...
+	native_output, err := native_resource.OrphanedResourceOperation(simpleorphanedsmithygeneratedtypes.OrphanedResourceOperationInput{SomeString: &someString})
+	if err != nil {
+		return Wrappers.Companion_Result_.Create_Failure_(err)
+	}
+	dafny_output := simpleorphanedsmithygenerated.OrphanedResourceOperationOutput_ToDafny(*native_output)
+	return Wrappers.Companion_Result_.Create_Success_(dafny_output)
+}
+
+func (CompanionStruct_Default___) CallNativeOrphanedError(input SimpleOrphanedTypes.Error) SimpleOrphanedTypes.Error {
+	native_error := simpleorphanedsmithygenerated.Error_FromDafny(input)
```
rishav-karanjit, line 35: Non blocking suggestion because Go is in TODO anyway: 

This is because `Error_FromDafny` returns `error` which is a interface type and should be asserted so that we can use `native_error.Message`.
```suggestion
	native_error := simpleorphanedsmithygenerated.Error_FromDafny(input).(simpleorphanedsmithygeneratedtypes.OrphanedError)
```

rishav-karanjit (approved):
LGTM. 

Had few non blocking comments. Those are non blocking because Go is in TODO.

PR #747, opened 2024-12-12
feat: partial @streaming blob support for Python

robin-aws:
Adds partial support for the `@streaming` Smithy trait on blobs, depending on a `actions-and-streaming-stdlibs` branch of Dafny that adds an Actions standard library, and only implemented when targeting Python so far. Event streaming is not technically supported yet because more codegen changes are necessary, but this will be a relatively small delta, as the Dafny types involved are already handled.

The motivation behind merging this incomplete work is largely to get the test model coverage into CI, so that it doesn't regress as the feature is completed, especially when Dafny verification is perturbed in future versions. The next priority will be to complete the Dafny branch and merge it to Dafny mainline.

Note that `smithy-dafny` will still emit a `DANGER` validation event saying that `@streaming` is not supported, which is intentional in order to guard against using the feature in production before it is fully completed: the test models exercising it explicitly suppress this event.

* Added a `dafny` submodule under `TestModels/dafny-dependencies/dafny` tracking the `actions-and-streaming-stdlibs` branch, just to access the contents of `Source/DafnyStandardLibraries/src/Std/Actions` etc. This copy of Dafny is never built or used for any other building tasks.
* Added a build task under `StandardLibrary` to build a .doo file from a subset of the Dafny standard library files above (using `DafnyStandardLibraries-smithy-dafny-subset.toml`). This is then used as a `--library` argument when verifying/translating polymorph libraries. This approach allows the standard library source to be built with different but compatible options such as `--reads-clauses-on-methods`.
  * This is guarded by a `USE_DAFNY_STANDARD_LIBRARIES` Makefile variable, since a newer Dafny is required for this to work.
* Bumped the high watermark Dafny version tested in CI to 4.9.1, which is necessary since the standard library above uses newer features like `@`-attributes.
* Modified Dafny and Python codegen to emit references to the `ByteStream` and `RewindableByteStream` types, and removed the previous workaround to first `.read()` all data in Python.
  * Because streams have mutable state, the Dafny code generation needs to take them into account when generating things like `modifies` clauses. I haven't yet implemented this, instead using a patch file to add the missing specification to the `Streaming` test model code.
* Modified the `aws-sdk/s3` test model and fleshed the `Streaming` test model out to actually test the feature.
* Removes not-grep, since our globs are actually completely broken and the tool doesn't support our current code layout, and it was complaining about the contents of the `dafny` submodule. See #763.

Reviewer note: it is worth reading the contents of the pending [Actions](https://github.com/dafny-lang/dafny/tree/actions-and-streaming-stdlibs/Source/DafnyStandardLibraries/src/Std/Actions) and [Streams](https://github.com/dafny-lang/dafny/tree/actions-and-streaming-stdlibs/Source/DafnyStandardLibraries/src/Std/Streams) Dafny libraries, but they are still under construction and I'm not claiming they will not change before being merged. Interface feedback is welcome but is not intended to be a hard requirement in scope for this review.


TestModels/Streaming/Model/SimpleStreamingTypes.dfy
```diff
@@ -0,0 +1,349 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
lucasmcdonald3, line 1: We don't normally commit this file, right? Is this intentional?

TestModels/Streaming/README.md
```diff
@@ -4,6 +4,7 @@ This project will implement the smithy trait [streaming](https://smithy.io/2.0/s
 
 ## Status
 
-This project does not build. The `software.amazon.polymorph.smithydafny` project does not support code generation for the "streaming" trait.
-
-Once the Polymorph code generator supports code generation for this trait, these files should be extended to complete the type implementation.
+This project only currently works on Python: the Dafny code generation
```
lucasmcdonald3, line 7: I'd also note that someone will need to copy-paste these from the TestModels' StandardLibrary to the MPL's StandardLibrary

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/smithy_dafny_standard_library/internaldafny/extern/streams.py
```diff
@@ -0,0 +1,58 @@
+
+from _dafny import Seq
+from smithy_python.interfaces.blobs import ByteStream
+from smithy_dafny_standard_library.internaldafny.generated.Std_Streams import ByteStream as DafnyByteStream, RewindableByteStream as DafnyRewindableByteStream
+from smithy_dafny_standard_library.internaldafny.generated.Std_Enumerators import Enumerator
+from smithy_dafny_standard_library.internaldafny.generated.Std_Wrappers import Option, Option_Some, Option_None
```
lucasmcdonald3, line 6: Odd -- in my local this is just `Wrappers`. Did something change?

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/smithy_dafny_standard_library/internaldafny/extern/streams.py
```diff
@@ -0,0 +1,58 @@
...
+      case 0:
+        position = offset
+      case 1:
+        position = self.dafny_byte_stream.Position() + offset
+      case 2:
+        position = self.dafny_byte_stream.ContentLength() + offset
+    return self.dafny_byte_stream.Seek(position)
+
+
+class StreamingBlobAsDafnyDataStream(DafnyByteStream):
```
lucasmcdonald3, line 53: Just for others' clarification maybe:

```suggestion
class StreamingBlobAsDafnyDataStream(DafnyByteStream):
  """Wrapper class allowing a native streaming blob to be accessed from Dafny streaming code."""
```

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/smithy_dafny_standard_library/internaldafny/extern/streams.py
```diff
@@ -0,0 +1,58 @@
...
+    # TODO: assert size is -1, buffer, 
+    # or define a more specialized Action<int, bytes> type for streams.
+    next = self.dafny_byte_stream.Next()
+    while next.is_Some and len(next.value) == 0:
+      next = self.dafny_byte_stream.Next()
+    # Do NOT return None, because that indicates "no data right now, might be more later"
+    return bytes(next.value) if next.is_Some else bytes()
+
+
+class RewindableDafnyByteStreamAsByteStream(DafnyByteStreamAsByteStream):
```
lucasmcdonald3, line 29:
```suggestion
class RewindableDafnyByteStreamAsByteStream(DafnyByteStreamAsByteStream):
  """Wrapper class allowing a Dafny byte stream to be accessed and used in native code as a rewindable byte stream."""
```

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/smithy_dafny_standard_library/internaldafny/extern/streams.py
```diff
@@ -0,0 +1,58 @@
+
```
lucasmcdonald3, line 1: I'm a little confused by the intention behind the implementation.
Should/could these methods extend the generated code (like a typical extern), or are should these methods be in separate classes and not linked to the generated code?
Right now, generated Dafny won't have these methods. But they could, if you want to set that up.

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/smithy_dafny_standard_library/internaldafny/extern/streams.py
```diff
@@ -0,0 +1,58 @@
...
+from _dafny import Seq
+from smithy_python.interfaces.blobs import ByteStream
+from smithy_dafny_standard_library.internaldafny.generated.Std_Streams import ByteStream as DafnyByteStream, RewindableByteStream as DafnyRewindableByteStream
+from smithy_dafny_standard_library.internaldafny.generated.Std_Enumerators import Enumerator
+from smithy_dafny_standard_library.internaldafny.generated.Std_Wrappers import Option, Option_Some, Option_None
+
+
+class DafnyByteStreamAsByteStream(ByteStream):
```
lucasmcdonald3, line 13: I added some comments to help me understand what's going on -- feel free to discard if these are already clarified elsewhere
```suggestion
class DafnyByteStreamAsByteStream(ByteStream):
  """Wrapper class allowing a Dafny byte stream to be accessed and used in native code as a byte stream."""
```

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/smithy_dafny_standard_library/internaldafny/extern/streams.py
```diff
@@ -0,0 +1,58 @@
...
+  
+  def __init__(self, dafny_byte_stream):
+    self.dafny_byte_stream = dafny_byte_stream
+
+  def read(self, size: int = -1) -> bytes:
+    # TODO: assert size is -1, buffer,
```
lucasmcdonald3, line 20: I might suggest adding an `__iter__` method to these classes so they're broadly recognized as streams.
Notably, Pandas will say these classes aren't "file-like" because they don't have an `__iter__` method: https://pandas.pydata.org/pandas-docs/dev/reference/api/pandas.api.types.is_file_like.html

lucasmcdonald3 (commented):
I mostly just reviewed Python.
Nothing blocking, just some suggestions/questions.

TestModels/Streaming/Model/SimpleStreamingTypes.dfy
robin-aws: I had to for now because I'm patching in some missing clauses

TestModels/Streaming/README.md
robin-aws: My plan after merging this is actually to merge those so that only the smithy-dafny copy exists

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/smithy_dafny_standard_library/internaldafny/extern/streams.py
robin-aws: All of the code under the Dafny standard libraries are namespaced under `Std::`, so this is that copy of Wrappers.dfy rather than the local one from `dafny-lang/libraries`

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/smithy_dafny_standard_library/internaldafny/extern/streams.py
robin-aws: Fair, I'll add some more comments in general in fact

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/smithy_dafny_standard_library/internaldafny/extern/streams.py
robin-aws: I think the comments you left clear this up (and I'll integrate and expand on them and add more comments at a wider scope too): one class wraps the Python streaming concept up with the Dafny-generated one, and the other wraps up the Dafny-generated one with Python's interface. So they are externs in both directions.

TestModels/Streaming/README.md
lucasmcdonald3: Gotcha. Let's chat when you do. Crypto Tools might need to update our release process.

robin-aws: For sure! I don't think it will change much, just that MPL components will depend on the StandardLibrary in the smithy-dafny submodule instead of it's own copy

TestModels/Streaming/Model/SimpleStreamingTypes.dfy
```diff
@@ -0,0 +1,349 @@
...
+  )
+  datatype BinaryOfOutput = | BinaryOfOutput (
+    nameonly binary: ByteStream
+  )
+  datatype ChunksInput = | ChunksInput (
+    nameonly bytesIn: ByteStream ,
+    nameonly chunkSize: CountingInteger
+  )
+  datatype ChunksOutput = | ChunksOutput (
+    nameonly bytesOut: ByteStream
```
seebees, line 27: So this is from `Std.Streams` right?
I'm wondering if it would be better to not open these modules? I get why, but coming in and reading the code I have to imagine what is going on.

TestModels/Streaming/src/SimpleStreamingImpl.dfy
```diff
@@ -0,0 +1,145 @@
...
+  {true}
+  function ModifiesInternalConfig(config: InternalConfig): set<object>
+  {{}}
+  predicate CountBitsEnsuresPublicly(input: CountBitsInput , output: Result<CountBitsOutput, Error>)
+  {true}
+
+  method CountBits ( config: InternalConfig , input: CountBitsInput )
+    returns (output: Result<CountBitsOutput, Error>)
+  {
+    var counter := new Folder<BoundedInts.bytes, int>(0, (sum, byte) => sum + BytesBitCount(byte));
```
seebees: lol, This is an unfortunate name. I thought it was a directory at first :)

TestModels/Streaming/src/SimpleStreamingImpl.dfy
```diff
@@ -0,0 +1,145 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+// Do not modify this file. This file is machine generated, and any changes to it will be overwritten.
+include "../Model/SimpleStreamingTypes.dfy"
+module SimpleStreamingImpl refines AbstractSimpleStreamingOperations {
```
seebees:
```suggestion
module {:options "/functionSyntax:4" } SimpleStreamingImpl refines AbstractSimpleStreamingOperations {
```

TestModels/Streaming/src/SimpleStreamingImpl.dfy
```diff
@@ -0,0 +1,145 @@
...
+
+  method CountBits ( config: InternalConfig , input: CountBitsInput )
+    returns (output: Result<CountBitsOutput, Error>)
+  {
+    var counter := new Folder<BoundedInts.bytes, int>(0, (sum, byte) => sum + BytesBitCount(byte));
+ 
+    ForEach(input.bits, counter);
+
+    // Should really have the Folder fail instead,
+    // but this is a simpler correct approach.
```
seebees: I agree. I would like to see the other approach. For a friend :)

TestModels/Streaming/src/SimpleStreamingImpl.dfy
```diff
@@ -0,0 +1,145 @@
...
+    assume {:axiom} input.bytesIn.history == [];
+    var chunker := new Chunker(input.bytesIn, input.chunkSize);
+    var chunkerStream := new EnumeratorDataStream(chunker, input.bytesIn.ContentLength());
+    
+    return Success(ChunksOutput(bytesOut := chunkerStream));
+  }
+
+}
+
+module Chunker {
```
seebees: I get why you put this here, but it should be in a different file

TestModels/Streaming/test/SimpleStreamingImplTest.dfy
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    method TestCountBits(client: ISimpleStreamingClient)
+      requires client.ValidState()
+      modifies client.Modifies
+      ensures client.ValidState()
+    {
+        var s: seq<BoundedInts.bytes> := [[0x0], [0x1, 0x2], [0x3], [], [0x4, 0x5]];
+        var e := new Enumerators.SeqEnumerator(s);
+        var stream := new EnumeratorDataStream(e, 5 as BoundedInts.uint64);
```
seebees: what is the 5?

TestModels/Streaming/test/SimpleStreamingImplTest.dfy
```diff
@@ -0,0 +1,54 @@
...
+        expect ret.sum == 7;
+    }
+
+    method TestBinaryOf(client: ISimpleStreamingClient)
+      requires client.ValidState()
+      modifies client.Modifies
+      ensures client.ValidState()
+    {
+        var s: seq<BoundedInts.bytes> := [[0x0], [0x1, 0x2], [0x3], [], [0x4, 0x5]];
+        var stream := new Enumerators.SeqEnumerator(s);
```
seebees: Are these just vestigial copy/paste?

TestModels/Streaming/test/SimpleStreamingImplTest.dfy
seebees, line 1: I'm not going to block,
but how do we get/pass errors?

There are also no Chunk tests.

I would also like some tests that could show that it really is a stream. Like have a sleep or something? Not sure.

TestModels/Streaming/test/WrappedSimpleStreamingTest.dfy
seebees, line 1: why only the one test?

TestModels/aws-sdks/s3/test/TestComAmazonawsS3.dfy
```diff
@@ -60,7 +67,13 @@ module TestComAmazonawsS3 {
         // we only care about the Body
         var MyBody := ret.value.Body;
         expect MyBody.Some?;
-        expect MyBody.value == expectedBody;
+
+        // TODO: These need to be generated as postconditions on GetObject instead
+        assume {:axiom} fresh(MyBody.value.Repr);
+        assume {:axiom} MyBody.value.Valid();
```
seebees, line 73: Nice having the targeted axioms! Thanks.
The only even more targeted we have found to be a little nicer is to put this in a `by` construction exactly where it is needed. e.g. `Collect() by {...` :)

seebees (commented):
Looking pretty good. Are you trying to get this in as is?

TestModels/Streaming/src/SimpleStreamingImpl.dfy
robin-aws: YUP, especially as it's main use is an example of an accumulator I'm definitely renaming it

TestModels/dafny-dependencies/StandardLibrary/runtimes/python/src/smithy_dafny_standard_library/internaldafny/extern/streams.py
robin-aws: As discussed since `__iter__` splits on newlines and I think there's no use for that, I'll leave this out for now.

TestModels/Streaming/test/WrappedSimpleStreamingTest.dfy
robin-aws: No good reason, failure to update the wrapped copy. Would love to find a way to not have to manually curate the set of tests to run in either runner.

TestModels/Streaming/test/SimpleStreamingImplTest.dfy
robin-aws: Yup! :)

TestModels/Streaming/test/SimpleStreamingImplTest.dfy
robin-aws: For errors in general I think the Action abstraction can handle anything you need, and I added a big explicit design callout on that: https://github.com/dafny-lang/dafny/pull/6074/files#diff-cb5145d4b584ea7656360f05c946523c48c5651815cea269b469a855844c3ed2R53 

For ByteStream specifically I think there does need to be errors in the stream of chunks somehow, and I added a small comment on that one. It will be a little tricky since a Smithy byte stream cannot produce modeled errors in the middle of the stream, only unmodelled things like ConnectionDropped etc. I suppose it will have to be an opaque `object` just like the service-wide error.

Added the missing Chunk test.

Demonstrating more stream-ness will need more Python adaptors or other languages to land, so I'll leave that for the next batch.

TestModels/Streaming/test/SimpleStreamingImplTest.dfy
robin-aws: Total count of all bytes, which (currently) must be known ahead of time. Added the parameter name at least.

TestModels/Streaming/src/SimpleStreamingImpl.dfy
robin-aws: See the link elsewhere for Errors in general.

TestModels/aws-sdks/s3/test/TestComAmazonawsS3.dfy
robin-aws: Good call. In these cases I'm going to put the energy into getting rid of them entirely instead though. :)

seebees (approved):
LGTM

lucasmcdonald3 (approved):

PR #748, opened 2024-12-13
fix(rust): validate config on client construction

alex-chew:
*Description of changes:*
In the Rust codegen, constraint validation is performed on operation calls instead of within structure builders' `build()` function. (The reasoning for this choice is documented in the description of #582.) But the validation wasn't applied to the constructors of `localService` clients, so attempting to construct a client with an invalid config could panic during conversion (in particular, when a `@required` field was missing).

This PR implements the missing validation in client constructors, and tests that both valid and invalid configs have the expected behavior when passed to the client constructor.


ajewellamz (commented):
LGTM

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/library/ModelTestCodegen.java
```diff
@@ -29,9 +30,32 @@
...
+ * and the test neither constructs the operation input nor performs the operation call.
+ * <p>
+ * Otherwise, if the {@code expect.success} property is present,
+ * the generated test asserts that no exception is thrown
+ * during the operation call and all setup steps
+ * (construction of client config, client, and operation input).
+ * <p>
+ * Otherwise, if the {@code expect.failure} property is present,
+ * the generated test asserts that an exception is thrown
+ * during either operation input construction or the operation call.
```
robin-aws, line 53: Clever solution!

The only issue is that `@smokeTests` is currently only implemented for Java (notice this file is under `.../smithyjava/...`), so this PR isn't actually testing the Rust support. And based on looking at the generated code I think you're missing the actual call to validate the input in `from_conf`.

I'd suggest adding a completely manual test or two to `Constraints/runtimes/rust/tests/simple_constraints_test.rs` to meet the testing bar, rather than taking on implementing `smokeTests` for Rust in this PR.

robin-aws (changes requested):

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyjava/generator/library/ModelTestCodegen.java
robin-aws: Scratch that, for some reason when I code generated locally I didn't see the new validation right away. And having the smoke test support include this for Java is awesome, I'm not suggesting removing it.

alex-chew: Thanks, and just to clarify, the manual Rust tests are already present: https://github.com/smithy-lang/smithy-dafny/pull/748/files#diff-d4d7fdb673501fddda845fbe959f8f559f69e314b8f77e77fb09293dbb49e81c

robin-aws: Realized that too after the fact, was looking in the wrong place

robin-aws (approved):

Issue #593, opened 2024-09-21
Add TestModel for reusing input/output structures


robin-aws:
This was particularly annoying for Rust, since the codegen for input and output structures we imitated from `smithy-rs` is quite different from other structures, but Dafny doesn't distinguish. The DB ESDK does this a ton but no other polymorph project or test model does.

robin-aws:
To clarify, the DB ESDK uses DDB input and output structures as members of other structures, but doesn't reuse them as direct input/output structures for any operations. The latter is what smithy-dafny currently explicitly forbids to make the implementation simpler (being able to traverse from a structure to the unique operation it is input or output for). 

The former is somewhat less problematic but still seems to go against current Smithy rules - a recent Smithy update started complaining about this and we worked around it by removing `@input` and `@output` from the DDB model.

PR #750, opened 2024-12-17
feat(rust): validate nested constraints

alex-chew:
*Description of changes:*
The Rust codegen previously didn't validated nested constraints - it would check the direct members of structures, but not any deeper. This PR implements nested constraint validation for all the supported aggregate types.

Some notes and open questions:

* The current implementation assumes that a (synthetic) operation input or output shape is only ever used by a single operation, but #593 suggests that may not be true in the DB ESDK. We should determine whether the assumption is safe, and if not, refactor `RustValidationGenerator` accordingly.
  * Robin's note: I ended up refactoring a bit to not assume this, if nothing else because the SimpleString test model already violates it.  
* The error messages only mention the particular member shape that failed validation. It would be more helpful to provide the full path to the failing shape, or at least the name of the member's container. The latter is probably just a matter of refactoring some templates/variables, but reused input/output structures might complicate that (see previous item).
* Since nested constraint validation isn't implemented in all of the runtimes (at least not in Java and .NET, AFAICT), I've added both positive and negative Rust tests, but only positive Dafny tests (to at least confirm this PR hasn't broken anything in other runtimes).
  * Robin's note: one of the positive tests ended up failing on Java as well because of #278, so I took the same sed-based approach to disable it only for Java. 

### DIFF
--- a/TestModels/Aggregate/runtimes/rust/src/lib.rs
+++ b/TestModels/Aggregate/runtimes/rust/src/lib.rs
@@ -14,6 +14,7 @@ pub mod error;
 pub(crate) mod implementation_from_dafny;
 /// All operations that this crate can perform.
 pub mod operation;
+pub mod validation;
 mod standard_library_conversions;
 mod standard_library_externs;
 pub mod types;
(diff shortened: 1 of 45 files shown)


robin-aws:
Pausing to avoid CI

robin-aws (approved):

PR #755, opened 2025-01-03
fix(codegen): Allow multiple paths for a shape

RitvikKapila:
*Issue #, if available:*

*Description of changes:*
Earlier there was only one path allowed for a shape. With this change, we allow multiple paths for a single shape.


ajewellamz (approved):
LGTM

PR #758, opened 2025-01-07
fix(Go): always generate code in an order

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*
We want smithy dafny generated code to have same order so that we can check git diff. To have same order, we need to avoid hashmap and hashsets. This PR sorts the top level shape before getting into shape visitor in a natural order (like A-Z for strings, or 1,2,3 for numbers). Once inside shape visitor, all the shapes will be sorted in the order they are getting visited.

Reference:
> This class makes no guarantees as to the order of the map; in particular, it does not guarantee that the order will remain constant over time.

https://docs.oracle.com/javase/8/docs/api/?java/util/HashMap.html

> It makes no guarantees as to the iteration order of the set; in particular, it does not guarantee that the order will remain constant over time.

https://docs.oracle.com/javase/8/docs/api/java/util/HashSet.html

Similar PR in for Smithy Dafny Java: **https://github.com/smithy-lang/smithy-dafny/pull/238**

*Testing*
Tested these changes in MPL on https://github.com/aws/aws-cryptographic-material-providers-library/tree/check-go-polymorph-diff. See the github action triggered from the latest commit on this branch (I made changes to workflow file to run CI to check diff when I commit on this branch)
 

ajewellamz (approved):
LGTM

PR #760, opened 2025-01-13
chore: support sync rust

ajewellamz:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/TestModels/Resource/runtimes/rust/tests/simple_resources_test.rs
+++ b/TestModels/Resource/runtimes/rust/tests/simple_resources_test.rs
@@ -1,5 +1,5 @@
-use simple_resources::types::simple_resource::SimpleResourceRef;
 use simple_resources::operation::get_resource_data::*;
+use simple_resources::types::simple_resource::SimpleResourceRef;
 use simple_resources::*;
 
 #[tokio::test]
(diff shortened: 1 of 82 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustLibraryShimGenerator.java
```diff
@@ -853,7 +853,7 @@ private RustFile operationOuterModule(
       variables.put(
         "operationSendBody",
         evalTemplate(
-          "$snakeCaseResourceName:L.inner.borrow_mut().$snakeCaseOperationName:L(input)",
+          "$snakeCaseResourceName:L.inner.lock().unwrap().$snakeCaseOperationName:L(input)",
```
robin-aws, line 856: I assume this works in non `--sync-rust` mode as well?

robin-aws (dismissed):
LGTM assuming we sort out the CI failures

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithyrust/generator/RustLibraryShimGenerator.java
ajewellamz: No. smithy-dafny only works in sync mode.
You told me to do that.

robin-aws: Yup I'm fine with that, I was just curious.

TestModels/Resource/runtimes/rust/tests/simple_resources_test.rs
```diff
@@ -22,13 +22,23 @@ async fn TestClient(config: SimpleResourcesConfig) {
 
 async fn TestNoneGetData(config: SimpleResourcesConfig, resource: SimpleResourceRef) {
     let input = allNone();
-    let result = resource.inner.borrow_mut().get_resource_data(input).unwrap();
+    let result = resource
+        .inner
+        .lock()
+        .unwrap()
```
robin-aws, line 28: Not blocking on this PR, but I think we should really have resource operations defined on the resource ref structs as well, so customers doesn't have to mess with internal things. `borrow_mut()` wasn't great before in terms of encapsulation and now it's a bit worse.

ajewellamz: Yes, that would be very helpful. Especially if we could then make `inner` private.
Should the be an issue or a sim or something?

robin-aws: https://github.com/smithy-lang/smithy-dafny/issues/771

robin-aws (approved):

PR #762, opened 2025-01-17
fix(utils): Make get orphan shapes to walk inside referred shape as well

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*


rishav-karanjit:
Closing this as of now. If shape is referred by referrence trait only then its considerred orphan as of now

PR #764, opened 2025-01-21
fix(go): use shape from the symbol properties in getType utils function

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*
`getType` function recursively moves symbol top to down to get the type of symbol but shape is always constant right now. This PR uses shape property from the symbol which will fix it.


ShubhamChaturvedi7 (approved):
LGTM

PR #765, opened 2025-01-21
fix(go): generate type conversion for resource shape with no operation

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*

Workflow in main-1.x:
- For each operation in resource shape:
    1. Visit its input and output for type conversion
    1. Check if the resource shape is not visited and visit it. 
        - In this step: ***If the resource shape does not have any operation its type conversion will NOT be created***

With this PR:
- For each operation in resource shape:
    1. Visit its input and output for type conversion
- Check if the resource shape is not visited and visit it. 

With this update: ***If the resource shape does not have any operation its type conversion will still be created.***

The only change in MPL we are expecting is the type conversion for resource shape gets created only after shapes inside input and output of the operation is visited. Test PR in MPL: https://github.com/aws/aws-cryptographic-material-providers-library/pull/1252

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/DafnyLocalServiceTypeConversionProtocol.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/DafnyLocalServiceTypeConversionProtocol.java
@@ -367,6 +367,7 @@ public class DafnyLocalServiceTypeConversionProtocol
                 );
               }
             }
+          });
             if (
               !alreadyVisited.contains(resourceShape.toShapeId()) &&
               resourceShape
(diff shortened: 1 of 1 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/DafnyLocalServiceTypeConversionProtocol.java
rishav-karanjit, line 1: Github is showing a lot of diff because of formatting changes. Better option: Use https://diff.corp.amazon.com/ to see the diff between  `rishav/go/fixresourceshape` and `main-1.x`

lucasmcdonald3 (approved):
LGTM

PR #769, opened 2025-01-22
fix(go): AWS SDK shape namespace

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*
Short description:
- return `<serviceName>` if shape in aws sdk models is inside input and output of operation
- add new parameter `model` to the smithyTypeNamespace function 

Long description:
All the shape which is input and output of operation is inside `github.com/aws/aws-sdk-go-v2/service/dynamodb` and all other inside those are in `github.com/aws/aws-sdk-go-v2/service/dynamodb/types`. However, in current `main-1.x`, smithyNameResolver return `<serviceName>types` (example: `dynamodbtypes` and `kmstypes`) in every scenario which is wrong. Updates from this PR loops in the model's operation shape and if the shape belong to input/output of the operation shape and uses either <serviceName>types or <serviceName> module for the shapes.

Note: We alias service/types module to  <serviceName>types. For example: `github.com/aws/aws-sdk-go-v2/service/dynamodb/types` is aliased into `dynamodbtypes` and `github.com/aws/aws-sdk-go-v2/service/kms/types` into `kmstypes` so that there won't be name collision.

Test PR in MPL after bumping smithy-dafny to this branch: https://github.com/aws/aws-cryptographic-material-providers-library/pull/1243
(No polymorph changes expected in MPL because none of the model in MPL depends on shape that are inside input and output of operation shape.)

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/nameresolver/SmithyNameResolver.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/nameresolver/SmithyNameResolver.java
@@ -7,9 +7,12 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Map;
 import software.amazon.smithy.aws.traits.ServiceTrait;
+import software.amazon.smithy.codegen.core.CodegenException;
 import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.model.shapes.OperationShape;
 import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.Model;
 
 public class SmithyNameResolver {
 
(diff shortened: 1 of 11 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/nameresolver/SmithyNameResolver.java
```diff
@@ -70,9 +75,18 @@ public static String smithyTypesNamespace(final Shape shape) {
...
       }
-      return sdkName.concat("types");
+      // Boolean to hold if shape is input or output of any operation
+      boolean isTopLevelShape = model
+        .shapes(OperationShape.class)
+        .anyMatch(op ->
+          op.getInput().filter(shape.getId()::equals).isPresent() ||
+          op.getOutput().filter(shape.getId()::equals).isPresent()
+        );
+      return isTopLevelShape ? sdkName : sdkName.concat("types");
```
ShubhamChaturvedi7, line 85: Why not overload the method and pass this boolean only when required?

rishav-karanjit: Because if we do that every time we call `smithyTypesNamespace` we will end-up doing a if-else check like below:
```
if (isShapeFromAWSSDK(shape)) {
    smithyTypesNamespace(..., boolean flag);
} else {
    smithyTypesNamespace(...);
}
```

ShubhamChaturvedi7 (approved):

PR #779, opened 2025-02-03
fix: Nightly build

robin-aws:
Rust can build with the latest nightly now (and the CI isn't building from source any more so it can't specify a commit any more)

Doesn't fully fix the nightly build but will once https://github.com/dafny-lang/dafny/issues/6090 is addressed.


ajewellamz (approved):
LGTM

Issue #780, opened 2025-02-03
EnumV2 support is inconsistent in AWS SDK codegen


kessplas:
In SmithyV2, enums can have [values](https://smithy.io/2.0/spec/simple-types.html#enum-values). 
Attempting to add ListObjectsV2 to the S3 module does not “just” work. The culprit is this enum definition: 

```
"com.amazonaws.s3#OptionalObjectAttributes": {
            "type": "enum",
            "members": {
                "RESTORE_STATUS": {
                    "target": "smithy.api#Unit",
                    "traits": {
                        "smithy.api#enumValue": "RestoreStatus"
                    }
                }
            }
        }
```

This enum breaks from the conventions followed by all other enums in the smithy-dafny AWS SDK ecosystem. The majority of enums, by convention, match their name and value (including all of the enums in KMS and DDB). Some enums diverge from this convention, but follow a more general rule; that the enum value may contain a non-standard delimiter which is replaced by the _ character, and thus match the name. For example:

```
        "com.amazonaws.s3#ServerSideEncryption": {
            "type": "enum",
            "members": {
                "AES256": {
                    "target": "smithy.api#Unit",
                    "traits": {
                        "smithy.api#enumValue": "AES256"
                    }
                },
                "aws_kms": {
                    "target": "smithy.api#Unit",
                    "traits": {
                        "smithy.api#enumValue": "aws:kms"
                    }
                },
                "aws_kms_dsse": {
                    "target": "smithy.api#Unit",
                    "traits": {
                        "smithy.api#enumValue": "aws:kms:dsse"
                    }
                }
            }
        },
```

Or as in ObjectCannedACL:

```
                "public_read": {
                    "target": "smithy.api#Unit",
                    "traits": {
                        "smithy.api#enumValue": "public-read"
                    }
                },
```

During codegen, the non-standard delimiter is replaced with _ and so the names and values match. 

In the case of RestoreStatus there is no delimiter so the names and values don’t match. This is a problem, because apparently the Dafny side of the AWS SDK codegen only ever uses the name, and the Python side only ever uses the (possibly edited) value. 

There might be a simple band-aid fix; the dumbest of which is to merely treat RestoreStatus as a special case and replace it with RESTORE_STATUS in the Python AWS SDK codegen. 
The better fix is to align use of names vs. values in general for AWS SDK codegen. So, what do other languages do? 

**Dafny**

`ComAmazonawsS3Types.dfy`

```
datatype OptionalObjectAttributes =
[...]

PR #781, opened 2025-02-03
chore: special case enum Dafny codegen for OptionalObjectAttributes, add ListObjectsV2 to S3 TestModel

kessplas:
*Issue #, if available:* https://github.com/smithy-lang/smithy-dafny/issues/780

*Description of changes:* This adds `ListObjectsV2` to the S3 TestModel, and a fix to allow for its correct generation. See the above issue for more details. This sidesteps the issue for the one enum we currently know about that induces the bug described therein (OptionalObjectAttributes/RestoreStatus).

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
@@ -345,11 +345,35 @@ public class DafnyApiCodegen {
     final EnumTrait enumTrait = stringShape
       .getTrait(EnumTrait.class)
       .orElseThrow();
+    if (enumTrait.getEnumDefinitionValues().get(0).equals("RestoreStatus")) {
+      System.out.println(enumTrait.getEnumDefinitionValues().get(0));
+      System.out.println(enumTrait.getValues().get(0).getValue());
+      System.out.println(enumTrait.getValues().get(0).getName());
+    }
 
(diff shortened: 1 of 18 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydafny/DafnyApiCodegen.java
```diff
@@ -349,12 +349,28 @@ public TokenTree generateEnumTypeDefinition(final ShapeId stringShapeId) {
     if (!enumTrait.hasNames()) {
       throw new UnsupportedOperationException("Unnamed enums not supported");
     }
+
+    // Dafny uses the enum's name, rather than its value here.
+    // Python (and possibly other runtimes) use values.
```
lucasmcdonald3, line 354: Do we know for sure that we should switch Dafny codegen from using names to using values, and not switch the runtime codegen from using values to using names?

kessplas: `ComAmazonawsS3Types.py` uses the value, not the name. As far as I know we can't change that within Smithy-Dafny, since it's generated by Dafny's `translate`. I'll also ask Robin about it, since his comment on https://github.com/smithy-lang/smithy-dafny/issues/780 implies that `smithy-dafny` prefers to use the name.

kessplas:
Closing in favor of https://github.com/smithy-lang/smithy-dafny/pull/784 which takes the name-first approach.

PR #788, opened 2025-02-19
fix: Dafny standard libraries .doo building fixes

robin-aws:
Fixes necessary to use this in the MPL as well: https://github.com/aws/aws-cryptographic-material-providers-library/pull/1321


kessplas (dismissed):
LGTM

PR #794, opened 2025-03-10
fix(java): empty nullable lists convert to optional empty lists

seebees:
null and .isPresent() of empty are not the same.

The AWS SKD has a specific way it handles these cases, and we are not adjusting that.

This converts java null => None, and [] => Some([]).

This means that for smithy-dafny
Java <=> Dafny conversions are lossless.


seebees:
Also see: https://github.com/awslabs/aws-encryption-sdk-specification/pull/290

ajewellamz (approved):
LGTM

josecorella (approved):

PR #797, opened 2025-03-12
fix(Go): Empty Blob fix

ShubhamChaturvedi7:
*Issue #, if available:*

*Description of changes:*
Adds the fix for empty blob, based on the test case here: https://github.com/aws/private-aws-encryption-sdk-javascript-staging/pull/768/files


rishav-karanjit (approved):
LGTM

PR #799, opened 2025-03-17
fix(Go): Encdoing for err and timestamps

ShubhamChaturvedi7:
*Issue #, if available:*

*Description of changes:*
Uses the encoded strings for Error messages and timestamps


lucasmcdonald3 (approved):
Code seems fine. How did you find this? Should we add tests?

ShubhamChaturvedi7:
I was looking at the type conversion layer diff and figured we are using strings for timestamps (and errors). We do have positive tests, and I have created a backlog for negative tests: https://issues.amazon.com/issues/CrypTool-5406

PR #800, opened 2025-03-25
chore(rust): optimize dafny runtime

ajewellamz:
*Description of changes:*

The changes in `TestModels/dafny-dependencies/dafny_runtime_rust/ `do not need to be reviewed, as they are simply copied from the latest `https://github.com/dafny-lang/dafny/tree/master/Source/DafnyRuntime/DafnyRuntimeRust`

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/AwsSdkToDafnyShapeVisitor.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/AwsSdkToDafnyShapeVisitor.java
@@ -122,7 +122,7 @@ public class AwsSdkToDafnyShapeVisitor extends ShapeVisitor.Default<String> {
     }
     return """
     func () %s {
-        var v []interface{}
+        v := make([]interface{}, 0, len(input))
         if %s == nil {return %s}
         for _, e := range %s {
         	v = append(v, e)
@@ -565,7 +565,7 @@ public class AwsSdkToDafnyShapeVisitor extends ShapeVisitor.Default<String> {
      var bits = math.Float64bits(%s%s)
         var bytes = make([]byte, 8)
         binary.LittleEndian.PutUint64(bytes, bits)
-     var v []interface{}
+     v := make([]interface{}, 0, 8)
      for _, e := range bytes {
       v = append(v, e)
      }
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/AwsSdkToDafnyShapeVisitor.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/AwsSdkToDafnyShapeVisitor.java
@@ -133,7 +133,7 @@ public class AwsSdkToDafnyShapeVisitor extends ShapeVisitor.Default<String> {
         dataSource,
         nilWrapIfRequired,
         dataSource,
-        someWrapIfRequired.formatted("dafny.SeqOf(v...)")
+        someWrapIfRequired.formatted("dafny.SeqFromArray(v, false)")
       );
   }
 
@@ -575,7 +575,7 @@ public class AwsSdkToDafnyShapeVisitor extends ShapeVisitor.Default<String> {
         nilCheck,
         dereferenceIfRequired,
         dataSource,
-        someWrapIfRequired.formatted("dafny.SeqOf(v...)")
+        someWrapIfRequired.formatted("dafny.SeqFromArray(v, false)")
       );
   }
 
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/DafnyToAwsSdkShapeVisitor.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/DafnyToAwsSdkShapeVisitor.java
@@ -133,10 +133,10 @@ public class DafnyToAwsSdkShapeVisitor extends ShapeVisitor.Default<String> {
       : dataSource;
     return """
     func () []byte {
-    b := make([]byte, 0, input.(dafny.Sequence).Cardinality())
     if %s == nil {
         return nil
     }
+    b := make([]byte, 0, input.(dafny.Sequence).Cardinality())
     for i := dafny.Iterate(%s) ; ; {
         val, ok := i()
         if !ok {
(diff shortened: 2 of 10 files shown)


codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/DafnyToAwsSdkShapeVisitor.java
```diff
@@ -132,20 +132,17 @@ public String blobShape(final BlobShape shape) {
...
-        if !ok {
-            return b
-        } else {
-            b = append(b, val.(byte))
-        }
+      func () []byte {
+      if %s == nil {
+          return nil
+      }
+      a := input.(dafny.Sequence).ToArray().(dafny.GoNativeArray);
```
rishav-karanjit: Should this be `%s.ToArray().(dafny.GoNativeArray)`?

rishav-karanjit: In Go, we only need assertion when a variable is interface (`input.(dafny.Sequence)`). When its not interface, we don't need assertion. And there are cases where the dataSource will be a interface and cases where it won't be. So, using datasource is safer.

So, I suspect `input.(dafny.Sequence)` would fail in MPL.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/DafnyToAwsSdkShapeVisitor.java
```diff
@@ -132,20 +132,17 @@ public String blobShape(final BlobShape shape) {
...
+    b := make([]byte, 0, a.Length())
```
rishav-karanjit: Just reminding myself that the default value when `b := []byte{}` and `b := make([]byte, ...)` is same. So, there should be no change in default value.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/SmithyToDafnyShapeVisitor.java
```diff
@@ -237,7 +237,7 @@ public String blobShape(final BlobShape shape) {
         dataSource,
         nilWrapIfRequired,
         dataSource,
-        someWrapIfRequired.formatted("dafny.SeqOf(v...)")
+        someWrapIfRequired.formatted("dafny.SeqFromArray(v, false)")
```
rishav-karanjit, line 240: Commenting for myself if I landed to this PR in future

`dafny.SeqOf` does not modifies v but `dafny.SeqFromArray` modifies v. However, we are fine if v is modified as v is not used after this.

rishav-karanjit (changes requested):

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/AwsSdkToDafnyShapeVisitor.java
```diff
@@ -565,7 +565,7 @@ public String doubleShape(final DoubleShape shape) {
      var bits = math.Float64bits(%s%s)
         var bytes = make([]byte, 8)
         binary.LittleEndian.PutUint64(bytes, bits)
-     var v []interface{}
+     v := make([]interface{}, 0, 8)
```
ShubhamChaturvedi7, line 568: Is Double always fit in 8 bytes?

rishav-karanjit:
Should the title of the PR be `chore(go): optimize dafny runtime`?

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/DafnyToAwsSdkShapeVisitor.java
ajewellamz: `b := []byte{}` creates an empty slice, so the concept of default value doesn't apply.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/AwsSdkToDafnyShapeVisitor.java
ajewellamz: 1) Yes, because it's explicitly 64 bits
2) It doesn't matter. I've created an empty slice with capacity for 8 bytes. If the result is more or less than 8 bytes, the correct result will still be given, just slightly less optimized.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/DafnyToAwsSdkShapeVisitor.java
ajewellamz: FYI -- the wrong version did fine in MPL and ESDK.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/awssdk/shapevisitor/DafnyToAwsSdkShapeVisitor.java
```diff
@@ -133,18 +133,15 @@ public String blobShape(final BlobShape shape) {
...
-        if !ok {
-            return b
-        } else {
-            b = append(b, val.(byte))
-        }
-    }
+      if %s == nil {
+          return nil
+      }
+      a := %s.(dafny.Sequence).ToArray().(dafny.GoNativeArray);
```
rishav-karanjit, line 139: The dataSource could also contain assertion. I know its very confusing so it was very hard to create a codegen for Go.

```suggestion
      a := %s.ToArray().(dafny.GoNativeArray);
```

rishav-karanjit: > The dataSource could also contain assertion.

Thats why in 131 it says `dataSource.startsWith("input.(")?...`

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithygo/localservice/shapevisitor/DafnyToSmithyShapeVisitor.java
```diff
@@ -287,18 +287,15 @@ public String blobShape(final BlobShape shape) {
...
-        if !ok {
-            return b
-        } else {
-            b = append(b, val.(byte))
-        }
-    }
+      if %s == nil {
+          return nil
+      }
+      a := %s.(dafny.Sequence).ToArray().(dafny.GoNativeArray);
```
rishav-karanjit, line 293:
```suggestion
      a := %s.ToArray().(dafny.GoNativeArray);
```

rishav-karanjit (dismissed):
LGTM.

Talked offline on slack, this PR will be merged with with `chore:...`

rishav-karanjit (approved):
LGTM

PR #804, opened 2025-05-20
feat(Python): Generate DDB item format conversion utility, small fixes

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

* Generate some logic to help convert between boto3 client- and resource-formatted shapes
* Small fixes that are needed to generate Python DBESDK:
  * Handle localService Config shapes defined in a different service namespace
  * Don't write a `from_dict` call on AWS SDK structures (these are already dicts)
  * `LocalServiceSymbolVisitor` delegates to `AwsSdkSymbolVisitor` sometimes (fixes DBESDK AttributeValue issue)
    * This triggered the `design-codegen.md` update. Ideally, there would be some AbstractSymbolVisitor that delegated to LocalServiceSymbolVisitor or AwsSdkSymbolVisitor when appropriate, but that's a larger refactor than is worth taking on, given that Crypto Tools isn't planning to develop Smithy-Dafny long-term
  * Handle no interceptors on config shape by setting default empty list


ajewellamz (approved):
LGTM

PR #805, opened 2025-06-09
chore(go): fix missing import 

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*
This PR adds the missing import in polymorph code. These import if missing will be added with Go imports, so there is no change in the generated code. However, before transpiling dafny code if we polymorph then goimports seems to miss some of the imports that is defined in dafny generated code like [here](https://github.com/aws/aws-database-encryption-sdk-dynamodb/actions/runs/15495124107/job/43629914666?pr=1936) where the codegen diff is failing.


ajewellamz (approved):
LGTM

PR #809, opened 2025-07-10
fix: smithy-python `OpaqueWithText`

imabhichow:
*Issue #, if available:*

*Description of changes:*
- Nested errors for SDK dependencies properly handle the `OpaqueWithText`
- see MPL PR for codegen diff and tests
https://github.com/aws/aws-cryptographic-material-providers-library/pull/1656

### DIFF
--- a/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/DafnyPythonLocalServiceProtocolGenerator.java
+++ b/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/DafnyPythonLocalServiceProtocolGenerator.java
@@ -450,6 +450,7 @@ public abstract class DafnyPythonLocalServiceProtocolGenerator
         );
         writer.addImport(".errors", "ServiceError");
         writer.addImport(".errors", "OpaqueError");
+        writer.addImport(".errors", "OpaqueWithTextError");
         writer.addImport(".errors", "CollectionOfErrors");
         writer.addStdlibImport("_dafny");
         writer.openBlock(
@@ -461,7 +462,7 @@ public abstract class DafnyPythonLocalServiceProtocolGenerator
               if error.is_Opaque:
                   return OpaqueError(obj=error.obj)
               elif error.is_OpaqueWithText:
-                  return OpaqueErrorWithText(obj=error.obj, obj_message=error.objMessage)
+                  return OpaqueWithTextError(obj=error.obj, obj_message=_dafny.string_of(error.objMessage))
               elif error.is_CollectionOfErrors:
                   return CollectionOfErrors(
                       message=_dafny.string_of(error.message),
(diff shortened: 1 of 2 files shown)


texastony (commented):
Did you test this fix with the MPL? Because if this the only fix we need, that's awesome.

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/DafnyPythonLocalServiceProtocolGenerator.java
```diff
@@ -461,7 +462,7 @@ private void generateErrorResponseDeserializerSection(
               if error.is_Opaque:
                   return OpaqueError(obj=error.obj)
               elif error.is_OpaqueWithText:
-                  return OpaqueErrorWithText(obj=error.obj, obj_message=error.objMessage)
+                  return OpaqueWithTextError(obj=error.obj, obj_message=error.objMessage)
```
imabhichow: > Did you test this fix with the MPL? Because if this the only fix we need, that's awesome.

I have tested this locally by applying the fix. Although the above fix is needed but it does NOT actually fix the below error
`AttributeError: 'Error_OpaqueWithText' object has no attribute 'message'`
It because nested errors for SDK dependencies do not properly handle the OpaqueWithText. For example, when a KMS error is nested with an OpaqueWithText error, our current code generation attempts to retrieve 'message' instead of 'obj_message'. I think the correct fix should be to attempt to resolve the nested error type and then attempt to extract the message.

texastony: OK. Do you have an estimate of how many days/hours it would take to run this down?

imabhichow: I think, based on my understanding from above initial fix, it would take a two or three days max to identify and come up with a solution and also test with MPL.( Or maybe it would take less than two days)
(cc. @lucasmcdonald3 may have more context on this)

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/DafnyPythonLocalServiceProtocolGenerator.java
```diff
@@ -679,7 +680,18 @@ private void generateErrorResponseDeserializerSectionForAwsSdkDependencyErrors(
...
+              elif hasattr(error.$L, "Message"):
+                  return $L(message=_dafny.string_of(error.$L.Message))
+              else:
+                  return $L(message=_dafny.string_of(error.$L.message))""",
+          code,
+          code,
+          code,
+          code,
+          code,
+          code,
```
texastony, line 694: This is outside the scope of this PR, but `code` is a bad variable name. 
I am going to assume it is a symbol for the Error Shape/symbol,
based on:
```
String code = AwsSdkNameResolver.dependencyErrorNameForService(
          context
            .model()
            .expectShape(serviceDependencyShapeId)
            .asServiceShape()
            .get()
        );
 ```

texastony (commented):
@seebees or @imabhichow is it possible for a Test model to test this?

texastony (approved):

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/DafnyPythonLocalServiceProtocolGenerator.java
imabhichow: >is it possible for a Test model to test this?

We synced offline and chose not to add a test for now. Since `smithy-dafny` is limited for us in the future.

PR #811, opened 2025-07-25
Align .NET converter generation scopes with local service/SDK scopes

### LINKED ISSUE #810: .NET generates conversions for a different scope of shapes than the other languages
See https://github.com/aws/aws-database-encryption-sdk-dynamodb/actions/runs/16479802597/job/46590838020?pr=1943#step:13:537. The `BucketSelectorReference` resource doesn't get picked up in `TypeConversionCodegen.findShapeIdsToConvert` when generating to .NET, but it is in scope for the other languages since they just filter to all shapes in the namespace.

robin-aws:
*Issue #, if available:*
Resolves #810

*Description of changes:*
Local services generally include orphaned shapes, whereas AWS SDKs generally only include shapes in the service closure. For most target languages, this means type conversion generation has to apply to a different set of shapes depending on whether the service is local or an AWS service. It turns out .NET was deviating from this by (mostly) using the same scope for both applications.

Resolved this by modifying the `TypeConversionCodegen.findShapeIdsToConvert()` logic to also include all convertable shapes in the service namespace, but maintained the original behaviour in the `AwsSdkTypeConversionCodegen` subclass.

Also fixed a bug in .NET this exposed where service references generated incorrect type names by ignoring `localService.sdkId`.

*Testing*

Modified `SimpleContraints` to have a different `sdkId` to expose the above bug in `Dependencies` (which had a significant ripple effect across languages in those test models).


ajewellamz (approved):
LGTM

PR #819, opened 2025-10-09
Optimize DafnySequence<Character> to String

olivergillespie:
Optimize performance of `ToNative$Simple.String`. I saw in an allocation profile of my application that this code is creating a lot of garbage, principally from calling `Character.toString` on every element, which creates a whole `String` object and backing `byte[]`. The `Stream` is wasteful too.

Move to a `StringBuilder` and for loop. A very rough benchmark shows about 80% reduction in allocations and total time taken.

Small string (4 chars):
| | Time | Allocation |
|---------|------|-------|
| Baseline | 159 ns/op | 608 bytes/op |
| Optimized | 37 ns/op | 96 bytes/op |

Larger string (88 chars):
| | Time | Allocation |
|---------|------|-------|
| Baseline | 2079 ns/op | 5360 bytes/op |
| Optimized | 419 ns/op | 816 bytes/op |

### DIFF
--- a/smithy-dafny-conversion/src/main/java/software/amazon/smithy/dafny/conversion/ToNative.java
+++ b/smithy-dafny-conversion/src/main/java/software/amazon/smithy/dafny/conversion/ToNative.java
@@ -64,7 +64,7 @@ public class ToNative {
     ) {
       StringBuilder sb = new StringBuilder();
       for (char c : dafnySequence) {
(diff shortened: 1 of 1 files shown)


olivergillespie:
Merged in https://github.com/smithy-lang/smithy-dafny/pull/820, thanks!

PR #826, opened 2026-02-25
 feat: support aws sdk .net v4 codegen

josecorella:
*Issue #, if available:*

*Description of changes:*
AWS SDK LakeFormation Service removed CREATE_TAG from the Permission Enum in the model:
https://github.com/aws/api-models-aws/blob/main/models/lakeformation/service/2017-03-31/lakeformation-2017-03-31.json


texastony (dismissed):

TestModels/aws-sdks/ddb/runtimes/net/tests/TestsFromDafny-cs.dtr
robin-aws, line 1: Should be omitted? (and same for other dtr files)

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/AwsSdkTypeConversionCodegen.java
```diff
@@ -66,9 +66,8 @@ public TypeConverter generateStructureConverter(
   }
 
   /**
-   * We can't call the {@code IsSet} methods on AWS SDK classes' member properties because they're internal.
-   * The best we can do is to call the properties' getters, which calls {@code GetValueOrDefault}, which in turn may
-   * improperly coalesce absent optional values to 0 (for example).
+   * In AWS SDK for .NET v4, value type properties are nullable and collection properties
+   * default to null. We extract optional members by checking for null.
```
robin-aws, line 70: Oof very good change to make to address that point then

codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithydotnet/TypeConversionCodegen.java
```diff
@@ -1026,27 +1051,9 @@ public TypeConverter generateUnionConverter(final UnionShape unionShape) {
               ) {
                 final TokenTree checkIfValuePresent;
 
-                // List<T> where T is not of type AttributeVale are always not null, but empty.
-                final Set<String> listTypes = Set.of("BS", "NS", "SS");
```
robin-aws, line 1030: So very happy every time we remove an "if DDB" hack :)

TestModels/aws-sdks/lakeformation/runtimes/net/Extern/LakeFormationClient.cs
robin-aws, line 1: Why did you have to add these files? IIRC they are generated, it's just that .gitignore might not recognize they shouldn't be checked in?

designs/AWS_SDK_V4_MIGRATION.md
robin-aws, line 1: This is very helpful, thanks for including it. Consider moving it to `designs` since it's not necessarily important enough for all users to see at the top level?

robin-aws (changes requested):

robin-aws (approved):

